PR #6, opened 2015-01-06
Update paths to inflammation files for the lesson 06-cmdline.

jdblischak:
I updated the last few instances where inflammation files were specified in the current working directory instead of in the subdirectory data/.  For the example of the exercise to write `find-pattern.R`, I simply changed the pattern since the inflammation files were no longer in the current working directory (listing the inflammation files via `list.files` would require passing two arguments). In my experience most novices already struggle with this lesson, so I did not want to make it any more complicated.

### DIFF
--- a/06-cmdline.Rmd
+++ b/06-cmdline.Rmd
@@ -181,9 +181,11 @@ The * is a wildcard character in the Unix Shell.
 Thus all the files in the current working directory are included as arguments to  arith.R.
 ```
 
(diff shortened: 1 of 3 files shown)


mlammens:
@jdblischak yeah, I can have a look at this either this evening or tomorrow.

mlammens:
@jdblischak Sorry for the delay. I reviewed this PR and it looks good; knited up nicely in my repo. One suggestion - at line 184 in 06-cmdline.Rmd, I would add a comment making it explicit that the pattern you are searching for is "print-args".

jdblischak:
Good idea, @mlammens. I added some more explanation. Thanks for the review.

PR #21, opened 2015-01-14
Add description of lesson content and also add advice for contributing to README

### LINKED ISSUE #10: Update README
Need to add brief description of lesson (see [Lesson Layout][layout]).

[layout]: https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md#lesson-layout

jdblischak:
Fixes #10.


jdblischak:
@dhaine, if you don't have time to deal with using the command line to resolve this, I can handle that if you review it first.

dhaine:
@jdblischak : maybe I missed a step when moving from old (bc) to new lessons, but do we always have to render the html files and push them? I understand if a new Rmd file is created, but if it's a modification to an existing file? Can the reviewer update the html and md files? Could be I just need to get to the week-end or can change "can render" to "have to render".

jdblischak:
Yes, we always have to render and push the html files. This is because in the new lesson template we are rendering the lessons using pandoc, instead of relying on the automatic jekyll support provided by GitHub (that's why the new repos have a file called ".nojekyll" in them; it prevents this automated building). You can read more about the design of the new setup [here][].

I'll update the language to make it more clear that this is required.

[here]: https://github.com/swcarpentry/lesson-template/blob/gh-pages/DESIGN.md

jdblischak:
@dhaine, how does this sound now? Are the directions clearer?

dhaine:
@jdblischak, good for me. Can you merge it?

PR #22, opened 2015-01-15
change example function name to not collide with base::outer function

pschmied:
`outer` is already defined in base. This just offered us the opportunity to caution about redefining built-in names for R, but it's probably best to not redefine built-in base R functions.

### DIFF
--- a/02-func-R.Rmd
+++ b/02-func-R.Rmd
@@ -112,10 +112,10 @@ fence(best_practice, asterisk)
 ```
 
   + If the variable `v` refers to a vector, then `v[1]` is the vector's first element and `v[length(v)]` is its last (the function `length` returns the number of elements in a vector).
-    Write a function called `outer` that returns a vector made up of just the first and last elements of its input:
+    Write a function called `firstlast` that returns a vector made up of just the first and last elements of its input:
     
 ```{r, include=FALSE}
-outer <- function(v) {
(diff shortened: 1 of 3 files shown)


dhaine:
Good catch @pschmied. Indeed `outer` is a built-in base function. But if you change the function name in the R lesson, you might want to change it also in the python lesson for compatibility between them. Maybe an easiest approach would not to change its name but add a note specifically in the R lesson warning novice users not to choose names already in use by built-in R functions?

jdblischak:
I think this is an example where it would be okay for the two lessons to diverge. For example, we use the variable name `dat` instead of `data` for this reason. But for the sake of keeping the two lesson sets easily comparable, I'd suggest creating a new function name that is closer to the original `outer`. How about `out` or `outside`?

pschmied:
I like "outside"! If others concur, I'll go ahead and amend my PR.

jdblischak:
@pschmied, please also render the lesson by running `make preview`. You'll need to install some dependencies first, which you can read about [here][].

[here]: https://github.com/swcarpentry/lesson-template#dependencies

pschmied:
Bueno,

Just went ahead and did those changes. Assuming I've done this correctly, this PR should have been updated.

jdblischak:
Thanks, @pschmied.

PR #23, opened 2015-01-16
Figure to fig

dhaine:


dhaine:
Changes /figure to /fig

jdblischak:
Great. This works for all the files in which we had explicitly set the name of the figures created. However, files which have not had this set will still save to `figure/` by default. I am thinking we should update [tools/chunk-options.R](https://github.com/swcarpentry/r-novice-inflammation/blob/gh-pages/tools/chunk-options.R) to save to `fig/` by default. Thoughts?

dhaine:
I believe it is comming from 04-cond-colors-R.Rmd. But is this lesson not in conflict with 04-cond.Rmd?

jdblischak:
I haven't removed 04-cond-colors-R entirely because someone may want to revive it (though that is obviously unlikely). Do you think we should just delete it?

dhaine:
We could either remove it or put it in "stand-by" for a certain period (6 months?) to see if anybody would want to use it or work on it. In the meantime we can switch off the production of plots (e.g. eval=FALSE) and add a comment at beginning of file.

jdblischak:
If we leave it there and edit it, I think it would be better to just go ahead and redirect the figures by adding the following line:

```
opts_chunk$set(fig.path = "figure/04-cond-colors-R-")
```

jdblischak:
Thanks, @dhaine.

Would it be difficult for you to set the language of R to English temporarily when you run `make preview`? The merge conflicts were due to error messages being printed in French. I left them in there for now because it will be easier for me to simply update with a new PR than manually fix all the error messages.

dhaine:
I'll see to switch it permanently to English to avoid future errors. Thanks!

Issue #19, opened 2015-01-14
Reformat challenges


jdblischak:
The challenges need to be reformatted to conform to the new lesson template, see [here][]. This could be tricky for those challenges that include R Markdown chunks. We'll probably just want to leave the chunks alone, but if someone thinks of a better idea that would be great.

[here]: https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md#topics

jdblischak:
Fixed by #27.

PR #27, opened 2015-01-21
Reformat challenges (#19)

dhaine:


jdblischak:
Thanks, @dhaine. This looks good. The only bad thing is that when the challenge includes knitr chunks, the remainder is formatted as a typical block quote and not as a challenge. I'm not sure if there is a way to improve this.

jdblischak:
Fixes #19.

dhaine:
I couldn't find a way to solve the post-chunk formatting issue. For now we can leave it like that until someone come with a better solution.

jdblischak:
Agreed.

PR #29, opened 2015-01-22
BUGFIX: modified list.files statement to exclude *.pdf files in data dir

tomwright01:
Not sure if this is the best fix, it might be better to move the pdf files.


jameshiebert:
That problem was bothering me as well... thanks for fixing it!

jdblischak:
Thanks for the PR, @tomwright01. I have been trying to figure out how to deal with the generated pdf files, too. While I like your solution, I would not want to try to explain `"inflammation-\\d+.csv` to a room of novices.

Here's another option. The pdf files get sent to `data` because the updated version of `analyze_all` in 04-cond.Rmd uses `full.names = TRUE` when retrieving the list of csv files with `list.files`. If we removed that argument, then the pdf files would be written in the base directory instead. Thoughts?

jdblischak:
And a related follow up question, should we commit the generated pdf files to the repo so that it would be easier to catch such bugs?

tomwright01:
The base directory is not a bad idea for learners who should be starting
with a clean directory structure. But could end up causing problems
later.
My preferred approach is to create another directory 'output/' and
introduce the paste command to generate the output path. This has the
advantages of 1) Introducing a 'very' useful command and 2) supporting
the link between what is happening in R and the file system. If R-Studio
is being used the new directory can be created directly from that
interface.

I realise introducing paste adds yet more complexity, perhaps the paste
function could be introduced back with the fence function in
01-starting-with-data. In my opinion the literal translation of the
python fence() function seems very artificial and would be improved by
adding paste() to 'pretty' the output.

jdblischak:
But is the tradeoff in complexity worth it? I've taught these materials and found that novices already struggle with just the basic idea of writing a function. Are you introducing these changes for your upcoming workshop? That would be a good way to test how learners handle the extra complexity.

tomwright01:
I don't think I did this correctly.
This was supposed to be a new PR with all the changes I made introducing factors, and more information on slicing dataframes.
I think I'm going to close this and see what happens when I push again. Sorry for cluttering peoples inboxes.

jdblischak:
@tomwright01, this was your previous PR with just the change to use a regular expression to exclude pdf files when using `list.files`. Since you submitted this PR on your gh-pages branch, when you updated your gh-pages branch with more commits, they were automatically added to this PR. This is one of the benefits of using feature (aka topic) branches, especially for PRs that may not be merged quickly.

PR #30, opened 2015-01-23
Fix bug with function `outside`

jdblischak:
Avoid error from overwriting function `outside` with the vector of the same name.

### DIFF
--- a/02-func-R.Rmd
+++ b/02-func-R.Rmd
@@ -230,9 +230,9 @@ That only works if functions don't interfere with each other; if they do, we hav
 >following:
 
 ```{r, results="hide"}
-inside <- "carbon"
-outside <- "+"
-result <- outside(fence(inside, outside))
+inner_vec <- "carbon"
+outer_vec <- "+"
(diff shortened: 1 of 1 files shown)


jdblischak:
@dhaine, could I please get a +1 on this?

dhaine:
+1

mlammens:
Should we be concerned that this overwrites/overloads the base function "outer"?

jdblischak:
Ugh...yeah, good catch, @mlammens. I had forget that is the reason we got changed the name of the function in the first place (see #22). How about this?

```
in <- "carbon"
out <- "+"
result <- outside(fence(in, out))
```

mlammens:
Good change @jdblischak , +1 from me.

dhaine:
Damn, forgot about that one!

jdblischak:
R doesn't allow the variable name `in`, so I went with more verbose variable names.

```
inner_vec <- "carbon"
outer_vec <- "+"
result <- outside(fence(inner_vec, outer_vec))
```

Issue #20, opened 2015-01-14
Reformat callout boxes


jdblischak:
The lessons currently have asides which start with **Tip:**. These need to be reformatted to conform with the new lesson template, see [here][].

To find all the occurrences of these asides, you can run `grep Tip: *Rmd`

[here]: https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md#topics

ateucher:
I've opened a pull request (#32) for this, but have just done the first file. You can wait to merge until I add more commits...

ateucher:
Closed by #32.

PR #32, opened 2015-01-27
Change Tips to callout boxes

ateucher:
This is a start on #20; reformatting the Tips as callout boxes.  You don't need to merge this yet, as I'll add more commits to it, but I thought I would push up changes to the first files to get the PR started.  

The figures were re-rendered when ran `make preview`, but are identical. Let me know if you don't want me to commit them.


jdblischak:
Thanks, @ateucher. I'll assign the issue to you.

No need to commit files that are modified, but not really changed. However, when there are new files you should commit those. For example, the newest version of knitr now adds a dash before the number when there are multiple plots generated by the same chunk, i.e. it used to be `chunk-plot2.png` but now it is `chunk-plot-2.png`.

jdblischak:
Well nevermind, I can't actually assign the issue to you. But since you referenced this Issue in this thread it should be fine.

gvwilson:
I've invited @ateucher to join the @swcarpentry team.

ateucher:
Thanks @gvwilson! I've assigned myself now.

ateucher:
@jdblischak I think this PR is ready for review and merge if it's acceptable.

jdblischak:
Looks good. Thanks, @ateucher.

Issue #18, opened 2015-01-14
Reformat learning objectives


jdblischak:
The beginning of each lesson has a list of learning objectives. This need to be reformatted to conform with the new lesson template, see [here][].

[here]: https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md#topics

PR #35, opened 2015-01-28
Converted objectives to new lesson-template (and a few other bonuses). Fixes swcarpentry/r-novice-inflammation#18

jameshiebert:


jdblischak:
Thanks, @jameshiebert. This looks good. The one thing I noticed is that you put the objectives above the chunk with all the knitr settings. I think it is important to have that chunk as close to the top of the file as possible since it explains to any potential contributor how we control the conversion to markdown. Could you please move the objectives below this chunk? Also, there is a second header that is explicitly written at the top of the R Markdown file that needs to be removed (now the title is generated from the setting in the YAML header). Could you please delete this header line as well?

jameshiebert:
Sure, no problem. I can make those fixes.

jdblischak:
Great. Thanks, @jameshiebert.

PR #36, opened 2015-01-28
Set fig.path in tools/chunk-options.R to send all plots to fig/

### LINKED ISSUE #11: Change figures/ to fig/
The subdirectory with images is currently `figures/`. This needs to be changed to `fig/`. Also the knitr chunk options at the beginning of each file needs to be updated to reflect this, e.g. `opts_chunk$set(fig.path = "figure/02-func-R-")` would be changed to `opts_chunk$set(fig.path = "fig/02-func-R-")`. See [Lesson Layout][layout] for more information.

[layout]: https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md#code-data-and-figures

jdblischak:
Fixes #11.

### DIFF
--- a/tools/chunk-options.R
+++ b/tools/chunk-options.R
@@ -6,7 +6,7 @@
 
 library("knitr")
 opts_chunk$set(tidy = FALSE, results = "markup", comment = NA,
-               fig.align = "center")
+               fig.align = "center", fig.path = "fig/")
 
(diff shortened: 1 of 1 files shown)


jdblischak:
Actually this is a bad idea. I should merge this into the core branch of lesson-template before merging into this repo.

PR #38, opened 2015-01-30
Remove intermediate Markdown files.

jdblischak:
For the new build process with pandoc, we no longer need the Markdown files. We edit the R Markdown files, and then running `make preview` knits the files, converts the Markdown to html, and then deletes the Markdown.


ateucher:
The only nice thing about the md files is that it's a quick and easy way to preview the files in the repo...

jdblischak:
Yes, good point, @ateucher. Greg and I had a similar [conversation][conv]. It is definitely a tradeoff.

The main pros of removing the md files are:

1. Reduces number of redundant files in an already large repository.
2. Our current setup using GNU Make automatically deletes these intermediate files.

The main pros of keeping the md files are:

1. Easy to quickly preview the rendered lesson from the GitHub repo.
2. Useful for understand what has caused merge conflicts. It's a lot harder to diagnose what the problem using only the html.

I'll send out an email to the mailing list to get feedback from other instructors.

[conv]: https://github.com/swcarpentry/lesson-template/pull/119

jameshiebert:
In my opinion, having the md files in the repo just adds more noise to the repository. When I'm looking over the changes created by a commit, I have to look and three times volume of what actually changed (Rmd, md, and html). I can live with having the html in the repo; I can see the value of having that there since it is the finished, polished lesson that is going to go online. But since the md files really are just and intermediary step, I'd rather not have them there.

I can see that it could be convenient to preview the rendered lesson from the GitHub repo, but I'd just as soon do it locally from my checked out repo. I don't have experience with pro number 2 in this context, so can't comment.

Thanks for considering this change.

jennybc:
I am a big fan of keeping intermediate Markdown precisely for the benefits on GitHub.

If someone wants to take a quick look at something, the immediacy of this is critical.

I think it's a mistake to assume someone will always be willing to clone an entire repo to their local machine.

jameshiebert:
I forgot to mention: if someone doesn't want to download the repo locally, they can always preview the final version on GitHub with [the repo's GitHub Pages](http://swcarpentry.github.io/r-novice-inflammation/) preview.

ateucher:
@jameshiebert That's true for the main lesson topics, and some, but not all of the supplementary lessons (though the links to these are in the text of the main lessons so take some work to find).

fmichonneau:
+1 to keep the markdown files. Having the rendered output directly
available in the repo next to the .Rmd files is really useful.

tomwright01:
The .md and .html files add a lot of noise when making lesson changes. Could they be added to .gitignore?

jdblischak:
@tomwright01, we can't ignore the html files. The html files are served via gh-pages to create the [website][] associated with the repo.

[website]: http://swcarpentry.github.io/r-novice-inflammation/

naupaka:
I agree with both sides of this argument.  What about if we put the different types of files for each lesson in different folders? For example, one folder for all the Rmd files, then one for md, and one for html.  Would be easy enough to do in the make script, I'd imagine, and then you have the benefits of using GitHub to preview the md files, plus the repo is cleaner. Thoughts?

stephenturner:
+1 to keeping .md files.

jainsley:
+1 to keeping them for all the reasons already mentioned.

jdblischak:
> I agree with both sides of this argument. What about if we put the different types of files for each lesson in different folders? For example, one folder for all the Rmd files, then one for md, and one for html. Would be easy enough to do in the make script, I'd imagine, and then you have the benefits of using GitHub to preview the md files, plus the repo is cleaner. Thoughts?

@naupaka, while we could automate the placement of files with the Make script, having all these files in different subfolders would add lots of complications. First, we want the paths to be as simple as possible since these lessons are for novices. If the Rmd files were in a subdirectory, the path to the data files would have to be `../data/` instead of the simpler `data/`. Also, the html files need to also be in the base directory because that is how they are served via the gh-pages branch.

jdblischak:
OK. Thanks for everyone that participated. There is more support for keeping the intermediate md files. Now we need to edit the Makefile so that it does not automatically delete these files. I'll open an Issue for the [lesson-template][temp]. If you're familiar with Make, please consider making the necessary edits.

[temp]: https://github.com/swcarpentry/lesson-template

PR #43, opened 2015-02-12
Challenge added: subsetting data

mikblack:
Added challenge for accessing and operating on subsets of data object.
Answer options cover potential misconceptions about R syntax.  
Alternative would be to provide numeric answers for each (3, 2.5, 1.5, 2) and ask students to write R commands (involving x and the mean command) that would produce those outputs.

### DIFF
--- a/01-starting-with-data.Rmd
+++ b/01-starting-with-data.Rmd
@@ -320,6 +320,20 @@ element[4:6]
 >
 > 1.  Use a slice of `element` to create a new character vector that spells the word "eon", e.g. `c("e", "o", "n")`.
 
+> ## Challenge {.challenge}
+>
+> Use R to calculate the mean of the first row of the following matrix:
+> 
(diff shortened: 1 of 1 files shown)


jdblischak:
Thanks for the PR, @mikblack. We purposely avoid the class matrix in the novice lessons so that we can focus on how to write code to perform a data analysis without getting bogged down in discussions of data structures. The challenge above involves slicing a vector. Perhaps you could add a challenge that involves slicing the data frame with the inflammation data, `dat`? For example, "Select the inflammation data for patients 3 through 6 on days 9 through 17."

gvwilson:
+1 to slicing the inflammation data frame - @mikblack can you please give it a throw?

mikblack:
Done - added and made pull request.

jdblischak:
Replaced by #49.

PR #46, opened 2015-02-13
fixing a small mixup of words

areejalsheikh:


01-starting-with-data.Rmd
```diff
@@ -356,6 +356,7 @@ The maximum value rises and falls perfectly smoothly, while the minimum seems to
 
 #### Key Points
 
+*
```
r-gaia-cs: Could you remove this line?

areejalsheikh: Actually, if you look through the pull requests of that file, I've added a new key point.

jdblischak:
Thanks, @areejalsheikh.

PR #52, opened 2015-03-04
Addressing data

tomwright01:
Added a supplemental lesson covering different ways of addressing data. This lesson is designed to stand alone but can follow the current lesson 01-starting-with-data.
It will be most useful in sessions where it is not required to cover the basic variables concepts in detail.


jdblischak:
I'd appreciate if you could remove the rendered html from this PR as well. This is a bit trickier because there are two separate rendering commits. If you are comfortable with interactive rebasing, that would be one way to fix it. Otherwise you can just roll back three commits and then re-add the names of the challenges.

Once you've done that, please send out a message to the r-discuss mailing list asking for a review of this supplemental lesson.

01-supp-addressingData.Rmd
```diff
@@ -0,0 +1,168 @@
...
+R is a powerful language for data manipulation. There are 3 main ways for addressing data inside R objects.
+
+* By index (slicing)
+* By logical vector
+* By name (columns and lists only)
+
+Lets start by loading some sample data:
+
+```{r readData}
+dat<-read.csv(file='data/sample.csv',header=TRUE, stringsAsFactors=FALSE)
```
naupaka: What is the general thought about using or not using `stringsAsFactors=FALSE` in these intro lessons?  I can see a number of cases in which is would be quite useful to leave it on.  I also worry that throwing this in there get's into a bit of a tangle, but I suppose the lesson materials are intended for the instructors to teach from and not for the students directly, in which case it doesn't matter all that much I guess.

01-supp-addressingData.Rmd
```diff
@@ -0,0 +1,168 @@
...
+A logical vector contains only the special values `TRUE` & `FALSE`.
+
+```{r logical_vectors}
+c(TRUE,TRUE,FALSE,FALSE,TRUE)
+```
+> ## Tip {.callout} 
+>
+> Note the values `TRUE` and `FALSE` are all capital letters and are not quoted.
+
+Logical vectors can be created using `relational operators` e.g. `<, >, ==, !=`.
```
naupaka: Also worth mentioning `%in%` here? Very useful and this would be the place to include it, if anywhere in the intro R materials.

01-supp-addressingData.Rmd
```diff
@@ -0,0 +1,168 @@
...
+>
+> Understand the 3 different ways R can address data inside a data frame.
+>
+> Combine different methods for addressing data with the assignment operator to update subsets of data
+
+R is a powerful language for data manipulation. There are 3 main ways for addressing data inside R objects.
+
+* By index (slicing)
+* By logical vector
+* By name (columns and lists only)
```
naupaka: Since this is just with data frames, is it necessary to mention lists here?

naupaka:
Nice lesson and certainly content that I think is important to have included for novices. I did notice that there is some overlap with the [existing lesson 1](https://github.com/swcarpentry/r-novice-inflammation/blob/gh-pages/01-starting-with-data.Rmd#manipulating-data). Not sure if this should go in there or if that should at the least have a pointer to this supplemental lesson. I also added a couple in-line comments.

01-supp-addressingData.Rmd
tomwright01: I added the stringsAsFactors=FALSE here because I really didn't want to worry about factors at the moment (I'm working on another supplemental lesson to address factors separately).  It becomes important for the final challenge. 
The text describing this command does comment on this, although I agree "don't worry about it..." may not be totally satisfactory.

01-supp-addressingData.Rmd
tomwright01: Great idea, I'll add that in.

01-supp-addressingData.Rmd
tomwright01: Agreed, perhaps another supplemental is called for ;-)

tomwright01:
There is overlap with lesson 1. It was unavoidable while still keeping this lesson modular. Personally I would like to include this content instead of the "Manipulating Data" section in lesson 1, but that approach makes lesson 1 pretty big and introduces a new dataset that messes up many of the functions introduced in lesson 1.

01-supp-addressingData.Rmd
dlebauer: I generally err on using defaults. It's easier to do this and perhaps introduce stringsasfactors after running into a gotcha (haven't tested out the lesson, but at first glance it seems everything will run as expected with factors). 

Also, why use `header = TRUE` if this is default?

My impression is that it is much easier

jdblischak: We don't currently have an official stance on the use of `stringsAsFactors = FALSE`. A few comments:

These lessons are not only for instructors. The reason we spend so much time building the lessons into nicely formatted html is so that self-directed learners can go through the online lessons themselves.

In general, I think it is ok to sometimes say "don't worry about this for now." With workshops only lasting two days, we have to focus on getting across the big picture.

I think you should remove `header = TRUE` since that is the default for `read.csv`.

01-supp-addressingData.Rmd
```diff
@@ -0,0 +1,174 @@
...
+The assignment operator `<-` can be combined with indexing.
+
+```{r indexing and assignment}
+x<-c(1,2,3,11,12,13)
+x[x < 10] <- 0
+x
+```
+
+> ## Challenge - Updating a subset of values {.challenge}
+> In this dataset, values for Gender have been recorded as both uppercase `M, F` and lowercase `m,f`.
```
jdblischak: This is also done in the lesson on factors you submitted. Is it your intention that only one or the other is taught?

tomwright01: Are you referring to PR #40 ?
My impression is that PR should be closed. I have added the new material from that into these two supplemental lessons PR #57 and PR #52 .

jdblischak: I am referring to line 162 in the lesson in PR #57. Is that not the same task? That is why I am confused by the order these should be taught. The answer to the challenge in this lesson is given a decent amount of discussion in the other lesson.

tomwright01: While the task is the same (update values from 'M' to 'm') the solution changes in PR #57 . The change from a character vector to a factor introduces a new step. The key new command in PR #57 is `droplevels()`.

Btw. Both lessons are designed to stand-alone. Personally, I would expect to teach addressing data before factors, but this is not a requirement.

PR #55, opened 2015-03-05
Modified readme.md to clarify what should be included in a PR

tomwright01:
Added pointer to CONTRIBUTING.md and added instructions specific to R lessons.


jdblischak:
Thanks for this, @tomwright01. Overall I like it a lot. I have one request for improvement. When you discuss `make preview`, you only mention that this is used for creating the html for display, and then right after say not to commit these files. While it is true that is how the html is built, I think you should additionally clarify that the reason a contributor should run `make preview` is so that they can view how their changes will be displayed in the final rendered version. As it reads now, it is not quite clear why a contributor would want to bother with installing the dependencies to build the html files.

jdblischak:
Great. Thanks, @tomwright01.

PR #56, opened 2015-03-05
 Add NA concept and other small modifications

fmichonneau:
* removed trailing spaces
* renamed/reformatted/restructured headers to improve structure and to make them
  more coherent with rest of document
* removed duplicated example
* add example for adding elements in front of vector
* add section about NA and test whether a vector contains NA
* attempt at clarifying section about coercion


fmichonneau:
Thanks for the feedback Tom! I included your suggestions.

jdblischak:
Thanks for the PR, @fmichonneau.

I've never quite known what to do with this lesson because it does not fit into the narrative of the analysis of the inflammation data. Also, it covers material that is covered later in the main lessons. What if we split this file into multiple supplemental files? The first parts on the basics of R and RStudio could be one file, e.g. `01-supp-intro-r-rstudio.Rmd`, and the sections on each of the data structures could be another, e.g. `01-supp-data-structures.Rmd`. I think the sections on indexing and functions can be removed because these subjects are covered in more detail in the main lessons. Also, the section on installing packages can be removed because no external packages are downloaded in these lessons. Thoughts?

jdblischak:
@fmichonneau, what are your thoughts on my suggestion? I'd like your opinion even if you don't have the time to implement the changes in this PR. If you don't have time, please let me know, and I'll merge in these changes.

fmichonneau:
Sorry for the delay. It has been a few busy weeks. I agree with your
suggestions, and I'll give it a shot at implementing them tomorrow.

fmichonneau:
OK, let me know what you think.

jdblischak:
Great! Thanks, @fmichonneau.

PR #57, opened 2015-03-05
Added topic on factors

tomwright01:
Factors are frequently encountered when using R. While they add significant functionality they often cause confusion. This supplemental topic tries to address some of the features and gotcha's associated with factors.
Much of the information is taken from the [datacarpentry](git@github.com:datacarpentry/datacarpentry.git) lesson on factors. The dataset is the same as in PR #52


01-supp-factors.Rmd
```diff
@@ -0,0 +1,189 @@
...
+f<-factor(c(3.4, 1.2, 5))
+as.numeric(f)
+```
+
+This does not behave as expected (and there is no warning).
+
+The recommended way is to use the integer vector to index the factor levels:
+
+```{r converting_factors_correctly}
+levels(f)[f]
```
naupaka: So THAT'S how you're supposed to do it.  Excellent.  I've always done as.numeric(as.character(f)). Which is clunky.

naupaka:
I think this is a great lesson. It should definitely be merged. Factors can be a huge source of confusion for novice and intermediate R users alike.

Not sure where it goes in the novice sequence, but I would guess that's not an issue at the moment since it's a supplementary lesson.

01-supp-factors.Rmd
```diff
@@ -0,0 +1,189 @@
...
+source('tools/chunk-options.R')
+```
+
+> ## Objectives {.objectives}
+>
+>   * Understand how to represent catagorical data in R
+>   * Know the difference between ordered and unordered factors
+>   * Be aware of some of the problems encountered when using factors
+
+This section is taken from the datacarpentry lessons git@github.com:datacarpentry/datacarpentry.git
```
jdblischak: It's not identical to the Data Carpentry lesson. Shouldn't you instead say that this lesson was "inspired by", "modeled after", or something to that effect? Also, it would be more useful to provide a URL to the lesson and not the ssh address for cloning the repo.

01-supp-factors.Rmd
jdblischak: @naupaka, I don't think this isn't a fair comparison of the clunkiness of the two methods. `levels(f)[f]` still returns a character vector. To obtain a numeric vector using this method, you'd have to run the following: `as.numeric(levels(f)[f])`. I don't see much difference in the clunkiness. Personally, I prefer `as.numeric(as.character(f))` because I can always remember it when I need to convert a factor to a number. The version with `levels` is recommended only because of a speed gain in specific use cases (see [here][so1] and [here][so2]).

@tomwright01, it's fine if you use the `levels` method, but please add `as.numeric` since the goal of this section was to convert factors to numbers.

[so1]: http://stackoverflow.com/questions/3418128/how-to-convert-a-factor-to-an-integer-numeric-without-a-loss-of-information
[so2]: http://stackoverflow.com/questions/6979625/arithmetic-operations-on-r-factors/6980780#6980780

tomwright01: @jdblischak updated, what silly behavior, perhaps a patch to levels() is called for.

jdblischak: I disagree. `levels` always returns a character vector. But this speaks to my point about the two strategies. With a quick glance at `as.numeric(as.character(f))`, it is clear what the code is doing. On the other hand, `as.numeric(levels(f)[f])` is more confusing at first glance. I for one rarely (possibly never) pass an index vector that is longer than the original vector to create a new longer vector, which is why I think `levels(f)[f]` looks so strange.

tomwright01: Personally, the `levels(f)[f]` phrase helped me unlock my understanding of factors.
Given that this is the recommended way to convert from a level I suggest we stick with it. I have separated the as.numeric() conversion to aid clarity in the process. I realize this negates any performance benefits but think this is a minor drawback.
It may also be worth pointing out that this is  a fairly artificial example, if the values in the vector are actually numeric it doesn't make sense to code them as a factor. The only time I come across this issue is when loading data from a file and a column has been incorrectly quoted. In this case it is often better handled with a colClasses argument to the read command.

Issue #58, opened 2015-03-08
Some parts needed to as.numeric before running function


chendaniely:
@mlammens taught at NYAS this week, and some parts needed an `as.numeric` cast.

This is a place holder to see what was different from the lesson in the repo and the lesson taught.

mlammens:
@chendaniely This problem occurred while going through the "Creating Functions" lesson. I deviated from the lesson by trying to apply the `center` function on a row of the `dat` data.frame as opposed to a column. This apparently violates recent conventions set for dealing with rows in data.frames. Though using `apply` with data.frames, where `MARGINS = 1` does not seem to have the same problems.

jdblischak:
Yes, this is an unfortunate aspect of the current lesson. The operations need to be done on columns, unless applying to all the rows with `apply`. One option to fix this would be to convert the data to a `matrix`, but that proposal was voted down in favor of keeping the data in the familiar `data.frame` object that novices are likely to encounter (see [PR][]).

We've already had a similar [discussion][]. It would be nice to have this point documented in the [discussion page][discuss] of this repo. Would one of you be able to send a PR?

[discussion]: https://github.com/swcarpentry/bc/issues/859
[PR]: https://github.com/swcarpentry/bc/pull/639
[discuss]: https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md#discussion-page

mlammens:
@jdblischak We don't have a discussion page setup in the repo yet, do we? Either way, I can submit a PR in the next day or two discussing this issue.

jdblischak:
Thanks for volunteering, @mlammens. You are correct that we do not currently have a discussion.md, it is issue #16. Please feel free to add information on this topic as well as anything else that is relevant to teaching this lesson in a workshop.

jdblischak:
Addressed by PR #61.

PR #61, opened 2015-03-12
Added DISCUSSION.md and note within to address issue #58

mlammens:
Created a DISCUSSION.md file for the R novice lessons. Within it I added notes to address issue #58


chendaniely:
@mlammens I forgot what you were trying to do during the workshop, but
`rowSums` and `rowMeans` will do rowwise `sum` and `mean` on dataframe

jdblischak:
Thanks for this, @mlammens. Here are some minor things to update:

*  Filename should be discussion.html
*  Change the header of the section to level 3 (i.e. `###`)
*  Mention that taking the mean/median of a row also needs to be avoided in lesson 01.
*  Mention that the recommended solution is to the problem is to not apply these functions to rows.
*  Remove the unnecessary spaces in the R code (see [here][adv-r] and [here][google]).

[adv-r]: http://adv-r.had.co.nz/Style.html
[google]: http://google-styleguide.googlecode.com/svn/trunk/Rguide.xml#spacing

mlammens:
@jdblischak RE your first point, should I be pushing the *.md or *.html file? I thought the former.

jdblischak:
Correct, you should push the md file. But you'll have to update the name of the html file in index.md as well.

mlammens:
@jdblischak I made the suggested changes.

jdblischak:
Thanks, @mlammens.

PR #62, opened 2015-03-22
Updating from template

gvwilson:
This brings in new CSS from the `styles` repo via `lesson-template` and recompiles the HTML.


jdblischak:
Looks good to me.

fmichonneau:
Looks good to me as well.

PR #63, opened 2015-04-02
Pull Request for R Lessons, Lesson 01-starting-with-data.Rmd  from Noush...

noushing:
Dear Software Carpentry R team, 

In this few lines, I am trying to add another useful feature of R. using the “()” around a statement will print its assigned value automatically.

Best,
Noushin

### DIFF
--- a/01-starting-with-data.Rmd
+++ b/01-starting-with-data.Rmd
@@ -145,6 +145,14 @@ weight_lb
 Since `weight_lb` doesn't "remember" where its value came from, it isn't automatically updated when `weight_kg` changes. 
 This is different from the way spreadsheets work.
 
+An alternative way to print the value of a variable is to use () around the assignment statement. Here is an example:
+
+```{r}
+(total_weight <- weight_kg + weight_lb)
+```
+
+The above command adds the values of `weight_kg` and `weight_lb`, assigns the result to the `total_weight`, and finally prints the assigned value of the variable “total_weight”.
+
 Now that we know how to assign things to variables, let's re-run `read.csv` and save its result:
 
 ```{r}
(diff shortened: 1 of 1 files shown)


jdblischak:
Hi, Noushin. Thanks for the PR. I apologize it took so long for me to respond; I recently taught a workshop which took up most of my SWC time.

Could you please provide some use cases for why this is essential for novices to learn? I rarely ever use this trick, and the only time I do is when I am writing teaching materials with knitr (since it saves me from having to retype the name of the variable). Also, I am concerned it will be confusing to novices since all the various syntax marks (`[`, `(`, and `{`) are so arbitrary. We want them to remember that parentheses are used for function calls.

Also, there is the concern about the length of the teaching materials. This lesson already contains more content than most instructors are able to cover in a typical workshop. See this [post][] on the discuss mailing list for a recent converstation about this issue.

[post]: http://lists.software-carpentry.org/pipermail/discuss_lists.software-carpentry.org/2015-March/002853.html

dhaine:
Agree with @jdblischak. Maybe it would be better as a Tip in 01-starting-with-data lesson?

noushing:
Hello John and Denis,

I've used this quick printing trick in programs and had been helpful, and
thought that mentioning it can have advantages for new learners as well.
However, I totally agree that learning about too many new things in the
introductions can distract students. Including it within the Tip, as Denis
suggested, is a good idea.

Thank you,
Noushin

jdblischak:
@noushing, do you intend to update the PR, including this as a Tip?

noushing:
Sure, I can add it as a Tip, and make another Pull Request for your
approval.

jdblischak:
You don't need to make a new PR. You can just update the current one. Any changes you make to your gh-pages branch will automatically be sent to this PR.

noushing:
Thanks.

jdblischak:
Thanks, @noushing.

PR #65, opened 2015-04-14
Using gapminder instead of inflammation data

aammd:
Hello Maintainer & everyone,

I've started editing the R lessons to use the [gapminder](https://github.com/jennybc/gapminder) dataset, rather than the inflammation dataset. I've taught with this dataset before; here are the advantages I believe it offers over inflammation.csv : 

* Easy to relate to - everyone can understand the importance of the variables in gapminder: GDP, life expectancy and population size are factors that influence everyone. 
* It has a header (ie  column names) - most scientists who are R users will only ever work with data that is labelled with column headings. In fact, in my observations, they usually have headers that are too long/too detailed, but **never** absent altogether
* It has categories - again, in almost every scientific application, at _some_ point, there are categories that need to be compared. In gapminder the categories are familiar to everyone: countries and continents. 
* It is big enough to make you _want_ to use a programming language, but small enough to be convenient. 
* It available from [CRAN](http://cran.r-project.org/web/packages/gapminder/index.html) which means that all you need is R and an internet connection to obtain it via `install.packages("gapminder")`. It is also on the internet at [tiny.cc/gapminder](tiny.cc/gapminder).

As of now I have only edited the first part of the first lesson. This PR is more of a means to start an interesting discussion on the topic (thanks @BillMills for the suggestion to do it this way). 

You'll also find a few other things I changed that are less radical, for example using Rstudio projects in place of `setwd()`

Looking forward to hearing what you think


stephenturner:
:+1:. I've been using a very slightly modified version of the gapminder data for teaching R, dplyr, ggplot2, etc. in my own workshops, using code that modifies and borrows extensively from SWC and Jenny Bryan's courses. It's worked very well for all purposes, and as Andrew mentioned, it's easy for pretty much anyone to relate to, not too big as to be cumbersome, but big enough to warrant exploring with dplyr and visualize over factors with ggplot2.

beroe:
This is good.  When you grab the CSV, it might also be mention how R can open a URL as if it were a file object, but maybe keeping it simple (and more similar to what they will usually doing, which is loading txt files) is best.

dhaine:
Can't look closely to it right now as I'm on the go but why not a fresh
r-novice-gapminder lesson?


*_____________________________________________Denis *Haine
+1-514-572-7174

2015-04-14 15:11 GMT-04:00 Stephen Turner <notifications@github.com>:

aammd:
@stephenturner that sounds awesome! Would you feel inclined to share your lessons here (and what is the best way to do that? I could add you as a collaborator on my branch -- unless there is a smoother way?)

aammd:
@beroe for sure, though as Jenny says (and i tend to agree) 
> One can also read data directly from a URL, though this is more of a party trick than a great general strategy.

```r
## data import from URL
gdURL <- "http://www.stat.ubc.ca/~jenny/notOcto/STAT545A/examples/gapminder/data/gapminderDataFiveYear.txt"
gdURL <- "http://tiny.cc/gapminder"
gDat <- read.delim(file = gdURL)
```

All the same it might be worth including as a "tip" or whatever, just in case an instructor drops their USB drive in a gutter and has to dl the data fast on somebody's computer

aammd:
@dhaine I think that would be a great idea (and more relaxing for me; I would not have to worry about keeping all the exact same content).
Are you The Maintainer?

karawoo:
There's been [talk on the R-discuss list](http://lists.software-carpentry.org/pipermail/r-discuss_lists.software-carpentry.org/2015-February/000276.html) previously about creating a new r-novice-gapminder lesson; I think @naupaka was planning on working on that but I don't know where it's at right now. I'm planning to use gapminder in an upcoming workshop so I'd be happy to compare notes/approaches with others once I see how that goes.

BillMills:
I just got off a debrief, where I saw that some work on a novice gapminder is already in the air; http://combine-australia.github.io/r-novice-gapminder/, originally from @sritchie73; @aammd, @naupaka and @karawoo, we need to combine our powers here, seeing as there's clearly a ton of demand and real effort going into this.

One question that needs to get asked: do we want both the inflammation and gapminder lessons for R? Every additional lesson is a big bump in maintenance overhead, and from my outside observer's perspective, gapminder seems like a pretty popular option - as @stephenturner points out, he (and I think others) have been using gapminder in lieu of inflammation for some time now. What does SWC-R think?

aammd:
@BillMills @sritchie73 WOW! That is frighteningly close to my vision for this lesson. Not exactly sure what to do now -- would it be better to simply fork http://combine-australia.github.io/r-novice-gapminder/ ? Do we keep this PR open?  When can that material be included in the SWC canon?

karawoo:
@BillMills thanks for linking to the existing lesson, that is very helpful. I'm curious if anyone who has taught it could speak to how long that material takes to teach? On first glance it looks like way more than could be covered in a single morning or afternoon session.

If maintaining separate lessons is too much work, and it sounds like it is, then I am in favor of switching to gapminder over inflammation. I know there were good reasons to use the inflammation data when translating the lessons from Python, and I appreciate all the work that people have put in to the inflammation lessons, but I think the gapminder data is more interesting and useful for an R lesson.

BillMills:
@karawoo +1 - but on the other hand, if people like the inflammation lesson and are happy maintaining it, then it's good to have two options. To clarify: keeping both might well absolutely be the right thing to do - I just want us all to check in once in a while to make sure we're all communicating about what we're doing in the field. I know over in Python town, the lessons said one thing and the instructors were doing another for quite some time, and no one said anything about it for ages :)

sritchie73:
@karawoo is right, there's far too much material in the existing repository to cover in a single day's worth of workshop. I think there's a lot of fat that could be cut out of the materials I put together (http://resbaz.github.io/r-novice-gapminder/), which is the reason I haven't submitted a pull request for those lessons (@aammd). 

I was planning to sit down with them (hopefully in the next month or so) to adapt them to something that more closely resembles the other Software Carpentry materials: at the moment it's missing the same kind of "story" through the lessons. I know that there has been a lot of work done over the last month for the r-novice-inflammation materials, so I suspect they're at a point where dropping in the gapminder dataset will be much easier than fixing up the existing r-novice-gapminder lessons.

fmichonneau:
I'm +1 on using the gapminder dataset.

I think it would be appropriate that it becomes the default teaching
dataset, but I wouldn't oppose creating an additional lesson if there is
enough interest to keep the inflammation lesson going.

naupaka:
I also am definitely in favor of using gapminder over inflammation, but don't object to keeping both. As far as the materials I have been working on/with, it's a bit more complicated than just updated R lessons, since I have also tried to adjust the shell and git lessons to work with the gapminder data as well, so the whole two days of a workshop fits together in a single flow. This is nice for teaching, but locks things into a particular order and would be pretty different from the versions of these lessons that would be taught in python workshop. So, for the purposes of updating the core R novice lesson, I think a simpler approach that is modular and focuses on just refocusing the existing lesson (at the current scope) around gapminder might make the most sense for now. I also really like the materials @sritchie73 made but agree that they are probably way too much for one or even two half day sessions on R.

chendaniely:
+1 for using gapminder and ditching inflammation.

From my understanding it is better suited for R's toolkit and use.
There was an initial debate whether we should be using matrices or dataframes in R when the R lessons were first being translated by @jdblischak.  We eventually went with dataframes because that's what R is used for.

The next problem with the inflammation dataset is that teaching the `apply()` part is strange.  Because of the way the dataset is setup, when we teach the `apply()` function, we naturally end up trying to look at average inflammation on an individual by individual basis, which means we'd be doing a row-wise apply on a dataframe.  You can't do that unless you first cast to a matrix.  This becomes a common point of confusion for our novice learners.

The R instructors had the pleasure of talking with @hadley and we talked about how the inflammation dataset is just not in the correct format to teach R with. @naupaka then brought up the fact that he was working on the current gapminder lessons for this purpose.

I think our R material has been developed enough that we can safely fork away from the python lessons.

aammd:
Thank you everyone who has contributed to this discussion so far! I will continue to work on the files in the PR. I am preparing for a workshop at [Simon Fraser University](http://ttimbers.github.io/2015-04-30-SFU/), organized by  @ttimbers .

I'm going to try to include content from @sritchie73 's awesome lesson plan, as well. 

if anyone wants to contribute directly to my fork of this, please let me know

And if anyone wants to correct my software-carpentry-github-flow-etiquette please do so!

BernhardKonrad:
This may not be the right place to bring this up, but as a reply to @eddelbuettel's email comment

> We are aware of the advantage of consistent instruction across
> languages (hence "inflammation everywhere") and we are also aware of the
> desire to teach to R's strength (hence "gapminder").  You can't have the cake
> and eat it.

Have there been thoughts to write a pandas version of gapminder lesson?

sritchie73:
@chendaniely I was also intending to use both the gapminder and inflammation datasets, to demonstrate both data frames and matrices (I tend to work with matrices a lot in R). I just didn't get time to incorporate the inflammation data into the data structures lesson.

griffinp:
I was one of the instructors at the workshop last week (along with  @nfaux and @hdashnow) where we used a modified version of Scott's materials (this one http://combine-australia.github.io/r-novice-gapminder/ ). 
I agree with all of the points @aammd  and @stephenturner made about the utility of the gapminder dataset for teaching R, and we would definitely use it again. 

*Side note: we were initially intending to revamp Scott's material using a more ecology+evolution-relevant dataset (the Owls dataset from Zuur et al. http://www.highstat.com/book2.htm) but eventually realised just how much work this entailed and reverted to gapminder!*

As Scott mentioned, there is still a lot of material in the gapminder lessons and they could definitely be shortened in places. For example, we didn't get to the dplyr section, and had to rush through control flow (unfortunately, because this is one of the most useful topics for beginners in my opinion). 
To briefly describe the modifications @hdashnow, @nfaux and I made from Scott's materials:

we **changed the lesson order** so that 
* we had an initial focus on doing things *without* external data, then learnt about reading data by importing gapminder, and used the gapminder data for the rest of the session. 
* we introduced ggplot plotting about midway through the session so that we could reinforce it in challenges in the later lessons

**shortened the sections on functions, vectorisation and control flow**
* there was a lot of technical stuff in here that, while interesting, we thought wasn't necessary for a mostly-beginner workshop

**included more challenges**
* in most of the sections

From discussions afterwards, we thought **it would be useful to**
* integrate the 'project management in Rstudio' theme throughout the entire session better (there were also some problems with packrat and built-in git not working for all versions of Rstudio and R so this needs a bit of troubleshooting)
* shorten the Data Structures section somehow and make it more interesting
* perhaps roll the Vectorisation section into Functions
* write more Advanced challenges (at least one per section) so the more experienced participants always have something to do

Finally thanks again @sritchie73 for all your work on the original material, much appreciated!

gvwilson:
Hi everyone,

Thanks very much for all of this - I've just mailed the r-discuss list with [a plan](http://lists.software-carpentry.org/pipermail/r-discuss_lists.software-carpentry.org/2015-April/000311.html) - please let me know what you think.

thanks again,
Greg

hdashnow:
Hi all, just to +1 @griffinp's comments about our experience with R-novice-gapminder

I think @sritchie73 has done a great job. I really enjoyed teaching it. It still needs a fair bit of trimming, but I already have lots of ideas about where that could occur.

I'd be willing to assist with maintenance for a separate R-novice-gapminder. I must warn you that I don't have heaps of spare time, so it would be more of a support role.

jdblischak:
Development of an R lesson that uses the gapminder data set will take place in the new repo [r-novice-gapminder](https://github.com/swcarpentry/r-novice-gapminder).

PR #66, opened 2015-04-20
add > for chunks within challenges

fmichonneau:
This is an initial draft, and I'd like some feedback before fixing all the blockquote elements in the lesson (cf. swcarpentry/lesson-template#211).

@r-gaia-cs submitted a feature request to knitr to allow chunks within blockquotes and @yihui implemented it promptly! :rainbow:  This is going to help us greatly and make our lessons prettier :sunrise: 

The only downside is that we can't include chunks that don't produce any visible output (i.e., `include=FALSE`) as they produce a blank line in the markdown that `blockquote2div` interprets as being the end of the challenge box.

Here I'm proposing to move these chunks before the challenge. Does anybody have a better suggestion?

PS: If you want to try this, you'll need the latest knitr from GitHub (1.9.23 or at least after https://github.com/yihui/knitr/commit/3f52d476c5dca3eae657cc1ec6174bd3216b510d)

Preview (with the older knitr verison)

screenshot1

Preview (with the latest knitr version and the content of this PR)

screenshot3


jdblischak:
This looks good, @fmichonneau. The only suggestion I have is that you should add some documentation for the answers now that they are moved before the challenge, e.g. something like

    ```{r challenge-answer-01, include=FALSE}
    # See below for challenge
    print_N <- function(N) {
      nseq <- seq(N)
      for (num in nseq) {
        print(num)
      }
    }
    ```

This way an instructor reading the lesson understands what these chunks are.

dhaine:
+1 for @fmichonneau proposition. The doc about challenge/answer structure could also be put into instructors.Rmd so you don't have to make a note about it in every lesson.

fmichonneau:
I believe I fixed all the chunks that were embedded into challenges.

I opted to use the option `echo=-1` and to have the function needed as the first element of the chunk so the chunks don't have to separated from the text/code for the challenge. There was only one place where it was not possible. Let me know if you also want me to submit the md/html pages.

jdblischak:
Great work, @fmichonneau and @r-gaia-cs! This looks great.

PR #70, opened 2015-04-21
01-starting-with-data.Rmd: e.g. -> i.e. in one of the challenges

kbroman:
> Use a slice of `element` to create a new character vector that spells the word "eon", e.g. `c("e", "o", "n").`

Rather than e.g. ("for example"), this should be i.e. ("in other words").


fmichonneau:
+1

jdblischak:
Thanks, @kbroman.

PR #71, opened 2015-04-22
Build site with chunks in challenges. Also remove rendered 00-first-timers

jdblischak:
@fmichonneau, please approve.


fmichonneau:
I had a quick look and it seems good.

It'll be easier once the pages are rendered to see if there are any issues.

Thanks!

PR #72, opened 2015-04-22
Added text to explain slicing more clearly in the intro lesson

bw4sz:
Test pull request to see if i understand the workflow. Completed for instructor training course.

I focused on this small example because its the first time we introduce slicing. It needed to be a bit clearer the order of syntax (df[row,column]) and where the was a comma. I also find it easier to write out in text the full explanation of what we are calling, rather than assuming the reader can understand the syntax.


jdblischak:
Thanks for the PR, @bw4sz. Please read the [contributing guidelines][contrib] specific to this repo. You'll need to edit the source Rmd file directly and submit only this file in the PR. And of course please let me know if you have any questions.

[contrib]: https://github.com/swcarpentry/r-novice-inflammation#contributing

PR #74, opened 2015-04-27
fixed formatting error, extra bullet point

sstevens2:


jdblischak:
Thanks for the PR, @sstevens2. For the R lessons, we write in R Markdown and then render to md and html. Could you please update your PR to change the Rmd file instead? See [here](https://github.com/swcarpentry/r-novice-inflammation#contributing) for contributing guidelines.

sstevens2:
Thanks @jdblischak.  Is it correct now?
Sarah

jdblischak:
This works. In the future though, you would only submit the changes to the Rmd (since this was such a small change it's okay that you also changed the md file). Thanks.

Issue #73, opened 2015-04-23
Use different variable name for slicing challenge


jdblischak:
In the first lesson, there is an [exercise on slicing][slice]. It is currently a bit confusing because the name of the vector to be sliced is `element`. This is awkward when explaining how to select specific elements of a vector. This should be updated to have a different variable name (and optionally the word in the character vector could also be changed to be consistent with the new variable name).

[slice]: https://github.com/swcarpentry/r-novice-inflammation/blob/gh-pages/01-starting-with-data.Rmd#challenge---slicing-subsetting-data-challenge

jdblischak:
Fixed by #79.

PR #79, opened 2015-05-02
changed character vector in slicing exercise to less confusing example

hgibling:
Saw this [issue](https://github.com/swcarpentry/r-novice-inflammation/issues/73) and decided to act on it. I did a practice lesson on this material and trying to figure out how to talk about elements of 'element' drove me crazy!

I chose an animal that was the same length and was also an isogram. It can even spell 'eon' :)


jdblischak:
Thanks, @hgibling!

Fixes Issue #73.

PR #82, opened 2015-05-04
My contributions to lessons 01 to 06

jgalgarra:
Reviewed lessons 01 to 06. Minor changes in lesson 1. General comment on compatibility of markdown versions and RStudio. From release v0.98.932 RStudio manages v2 markdown. This fact has at least two undesirable side effects.

* The .md files are deleted by default. To keep them it is necessary addind these lines in the .Rmd header
``` keep_md: true ```

* When there is an error inside a chunk of R code, knitr aborts. It is necessary adding this to the chunck header:
``` {r, error = TRUE} ```

There is a problem with the file names of the images generated by knitr. I have tested in three different environments and it always adds the sequence number to the end of the name file wihtout the "-" character. For instance, when I knit, ``` 03-loops-R-inflammation-011.png ```` instead of ``` 03-loops-R-inflammation-01-1.png ```. I have seen in the reporistory figures with both schemas, I haven't found any hint on the ```opts$chunk``` help to change this behaviour. I could change the .md files by hand, but this could mislead the following contributor.


jdblischak:
Thanks for the PR, @jgalgarra.

We do not use R Markdown v2 to build our lessons. We first render them to Markdown using knitr and then convert to html using pandoc. To build the lesson locally, you can run `make preview`. However, you only need to submit your changes to the Rmd files in your PR. Please see the contributing [guidelines](https://github.com/swcarpentry/r-novice-inflammation#contributing). Also, since you are interested in the build process, please read about the [design](https://github.com/swcarpentry/lesson-example/blob/gh-pages/DESIGN.md) of the build process and the [layout](https://github.com/swcarpentry/lesson-example/blob/gh-pages/LAYOUT.md) of the lesson content.

As for the names of the output files, these indeed are a pain. The naming scheme seems to change based on the version of knitr. Since you only need to submit the Rmd file, you do not need to worry about how your version of knitr names the files. But if you'd like to dive into what is causing this difference in naming, I'd be interested in learning more.

jgalgarra:
Hi John,

I understand the process, but the problem is that if someone clones locally
the directory and tries to
run the Rmd files with a fresh installation of RStudio and knitr will get
an error, because of these compatibility issues. Knitr has changed the
default configuration values, and now the .md files are deleted by default
and
the script crashes when finds an error.

The changes I introduced are just to avoid this crashes and are backwards
compatible.

Javier


2015-05-05 5:22 GMT+02:00 John Blischak <notifications@github.com>:

jdblischak:
I liked the addition of `error = TRUE` for the code chunks where we know the code is going to produce an error. Could you please resubmit a PR with only these changes to the Rmd files?

jgalgarra:
I reopen the PR witth the error=TRUE added in chunks of lesson 02 when necessary

jgalgarra:
I close this PR and open PR105 to merge these changes with thos described in PR 85

Issue #77, opened 2015-04-30
suppy "ignore.case=TRUE" in callings of "sub" function 


HKYang:
I suggest to suppy "ignore.case=TRUE" in callings of "sub" function in the section entitled "Making choice" within R course material.

The R function "sub" is used there to substitute the filename ending "csv" with "pdf". In Windows platforms, since file names are case insensitive, users are free to use capital letters. Unfortunately, pattern strings in "sub" function are case sensitive by default. So, it is necessary to accomodate the scenario of having capital letters in file name extension by supplying the argument value of "ignore.case=TRUE".

jdblischak:
Thanks for the suggestion, @HKYang.

Have you run into issues with this in a workshop?

In this specific instance, I don't think adding this argument is necessary. All the files use all lower case letters, and the students do not create the names themselves. They download the data as is, so I would be surprised if this were to cause an issue in a workshop.

In a bigger picture though, do we want to encourage Windows users to name their files in such a way? If they do want to name their files ".PDF", shouldn't the code they write explicitly reflect this, e.g. `sub("CSV", "PDF", f)`? I don't think we would want to encourage them to use a system of data management where they don't know what their filenames are.

Please let me know I misunderstood your suggestion.

HKYang:
Hi, John:
I agree with you that in workshops very unlikely this issue will be popped
up.
However, this is more about fault-tolerance in good programming practices
that we are supposed to teach in addition to programming skills.
Thanks for your attention.
-kai

PR #86, opened 2015-05-08
suggest an enhancement in course material

HKYang:
suggest an enhancement in course material

### DIFF
--- a/2015-04-29-kai-yang-CourseMaterialFix.md
+++ b/2015-04-29-kai-yang-CourseMaterialFix.md
@@ -1,0 +1,16 @@
+---
+date: 2015-04-29
+round: Round 12
+title: A fix in course material
+author: Kai Yang
+permalink: /2015/04/kai-yang-fix-on-course-material-R/
+tags:
+  - Fix
+  - Bug
+  - Lesson
+  - R
+---
+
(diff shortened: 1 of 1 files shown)


jdblischak:
Hi, @HKYang. We had discussed this issue in #77. I still don't think we should be encouraging scientists to be unsure how they name their data files. If they save their data files as "csv", the code should only read in files that end with "csv". And vice versa, if they save their data as "CSV", the code should only read in files that end with "CSV". I don't see this as a fault tolerance problem. They are not writing code that is supposed to read in any arbitrarily named data file, but only specific files for their project.

Also, for a PR you should normally just make the changes that you want to see. We don't need a Mardown file documenting what you propose to change. See https://github.com/swcarpentry/r-novice-inflammation#contributing for more details.

If you strongly disagree with me, please feel free to post to the [r-discuss][] mailing list to get others' opinions.

[r-discuss]: http://lists.software-carpentry.org/mailman/listinfo/r-discuss_lists.software-carpentry.org

HKYang:
Hi, John:
I think this issue is not about how we do file naming. It is more about
fault tolerance. Sometimes, we cannot control external environments. That's
what we cared when I was employed at Microsoft.
Thanks for discussion.
-kai

jdblischak:
I'm closing this PR for the reasons I detailed before. Please feel free to gather support from other R instructors to implement this change.

PR #90, opened 2015-05-13
Correct format for 02-func-R lesson

dhaine:


jdblischak:
Let's not get rid of all the subheadings. This will make it very hard to read this lesson.

https://github.com/swcarpentry/lesson-template/issues/220

PR #95, opened 2015-05-13
Correct format for 06-cmdline lesson

dhaine:


dhaine:
Sorry, had rename the old lesson to 05-cmdline this afternoon. Everything should be ok now.

PR #104, opened 2015-05-13
Correct format for 05-testing lesson

dhaine:


jdblischak:
This lesson isn't really usable for use at a workshop. There are no examples, and it reads like a textbook not a lecture. I propose we delete it and change 06-cmdline.Rmd to 05-cmdline.Rmd. Thoughts?

dhaine:
Alright for me! I'm removing it and renaming 06-cmdline. I'll check where links to it might be in the other lessons and make the changes.

PR #105, opened 2015-05-14
My contributions (former PRs 82 & 85)

jgalgarra:
I create a new pull request with the changes to lesson 01 described in PR 85 and adding {r, error = TRUE} for knitr default behavior in new releases as described in PR 82.


jdblischak:
Thanks, @jgalgarra.

PR #116, opened 2015-05-20
Update Lesson 6: Best Practices

mcdickenson:
- Fix missing wording in point #5
- Add explanation of home directory
- Add warning about side-effects in `source`d files

### DIFF
--- a/06-best-practices-R.md
+++ b/06-best-practices-R.md
@@ -41,11 +41,11 @@ The following error message indicates that R has failed to set the working direc
 Error in setwd("~/path/to/working/directory") : cannot change working directory
 ```
 
-Consider using the convention that the user running the script should begin in the relevant directory on their machine and then use relative file paths (see below). 
+Note that using `~` for the home directory is a convention on Unix systems and will allow others to run your code more easily on their machines than if the path specified your username.  Consider using the convention that the user running the script should begin in the relevant directory on their machine and then use relative file paths (see below).
 
 4. Use `#` or `#-` to set off sections of your code so you can easily scroll through it and find things.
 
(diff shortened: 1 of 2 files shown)


jdblischak:
Thanks for the PR, @mcdickenson. Could you please apply your changes to the source Rmd file? See https://github.com/swcarpentry/r-novice-inflammation#contributing

06-best-practices-R.md
```diff
@@ -41,11 +41,11 @@ The following error message indicates that R has failed to set the working direc
 Error in setwd("~/path/to/working/directory") : cannot change working directory
 ```
 
-Consider using the convention that the user running the script should begin in the relevant directory on their machine and then use relative file paths (see below). 
+Note that using `~` for the home directory is a convention on Unix systems and will allow others to run your code more easily on their machines than if the path specified your username.  Consider using the convention that the user running the script should begin in the relevant directory on their machine and then use relative file paths (see below).
```
jdblischak: I don't think this extra sentence explaining the tilde is necessary for two reasons:

1. The prerequisite for this lesson is an understanding of how to navigate a Unix filesystem.
2. The point is to use relative paths instead of absolute paths. While using a tilde is better than specifying the name of your home directory, it would still force others to put the directory in the exact same place relative to their home directory.

mcdickenson: Thanks for your feedback. I've removed that sentence.

mcdickenson:
Thanks @jdblischak. I've applied the changes to the Rmd file--sorry about that.

jdblischak:
Great. This is ready to merge.

We've currently [frozen development temporarily][freeze] as we prepare the lessons for release. I'll merge once the freeze has been lifted.

[freeze]: http://lists.software-carpentry.org/pipermail/maintainers_lists.software-carpentry.org/2015-May/000081.html

jdblischak:
Thanks, @mcdickenson.

PR #119, opened 2015-05-24
Minor updates to R lesson 1

emilliman5:
Added a tip about 1-based indexing in R vs. 0-based. Made some minor formatting changes

### DIFF
--- a/01-starting-with-data.Rmd
+++ b/01-starting-with-data.Rmd
@@ -38,7 +38,7 @@ To do all that, we'll have to learn a little bit about programming.
 
 ### Loading Data
 
-To load our inflammation data, first we need to tell our computer where is the file that contains the values. We have been told its name is `inflammation-01.csv`. This is very important in R, if we forget this step we'll get an error message when trying to read. We can change the current working directory using the function `setwd`. For this example, we change the path to the working directory where our scripts are stored that is named `swc`:
+To load our inflammation data, first we need to tell our computer where is the file that contains the values. We have been told its name is `inflammation-01.csv`. This is very important in R, if we forget this step we'll get an error message when trying to read the file. We can change the current working directory using the function `setwd`. For this example, we change the path to the working directory where our scripts are stored that is named `swc`:
 
 ```{r, eval=FALSE}
 setwd("~/swc")
@@ -208,6 +208,10 @@ dat[1:4, 1:10]
 The [slice](reference.html#slice) `1:4` means, "Start at index 1 and go to index 4."
 
 The slice does not need to start at 1, e.g. the line below selects rows 5 through 10:
+
+> ## Tip {.callout} 
+> R indexes are 1-based, not 0-based. That means that first element has an index of 1 instead of 0.
+
 
 ```{r}
 dat[5:10, 1:10]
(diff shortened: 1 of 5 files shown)


dhaine:
Thanks @emilliman5 for the fixes, good catches!

01-supp-read-write-csv.Rmd
```diff
@@ -17,7 +17,7 @@ source('tools/chunk-options.R')
 The most common way that scientists store data is in Excel spreadsheets. While there are R packages designed to access data from Excel spreadsheets (e.g., gdata, RODBC, XLConnect, xlsx, RExcel), users often find it easier to save their spreadsheets in [comma-separated values](reference.html#comma-separated-values-(csv)) files (CSV) and then use R's built in functionality to read and manipulate the data. In this short lesson, we'll learn how to read data from a .csv and write to a new .csv, and explore the [arguments](reference.html#argument) that allow you read and write the data correctly for your needs.
 
 
-### TRead a .csv and Explore the Arguments
+### Tread a .csv and Explore the Arguments
```
tomwright01: Tread?

01-starting-with-data.Rmd
```diff
@@ -209,6 +209,10 @@ The [slice](reference.html#slice) `1:4` means, "Start at index 1 and go to index
 
 The slice does not need to start at 1, e.g. the line below selects rows 5 through 10:
 
+> ## Tip {.callout} 
+> R indexes are 1-based, not 0-based. That means that first element has an index of 1 instead of 0.
```
tomwright01: Since nothing else in R is 0 based is this really necessary?

jdblischak: I agree this is confusing. This lesson is targeted at novice programmers. One nice thing about R is that we do not have to spend time discussing 0-indexing. Why would novices expect it to be any other way?

If we do want to make note of it, we should at least give them the context: explain that some other languages, e.g. Python, start indexing from zero.

emilliman5:
My thought was that an experienced Python/Perl/... programmer can still be a novice R programmer. I agree more context should be given as to why this matters.

dhaine:
Sorry, `Tread`should have been `Read`.

dhaine:
I have mixed feelings on the second point. At first I got the same opinion than @emilliman5, i.e. we don't really have totally complete novice participants, They would have (some) knowledge of something else and a single note about 0/1-index without context would be good enough. Now not sure if better to add context or just not talk about it (would prefer the latter). Do you want me to revert the merge?

jdblischak:
While it is true that workshop attendees come in with a variety of backgrounds, this lesson was written explicitly with a novice programmer in mind. This is evident throughout. For example, when introducing functions, it spends a large amount of time motivating the use of functions and explaining the idea of encapsulation. Similarly, the loops section starts with the motivations to use loops. An experienced Python/Perl/... programmer already knows why functions and loops are useful, and this lesson would be an inefficient means for them to pick up R syntax.

If we do keep the tip, it needs to have context. For example, it could say something like "R indexes start counting at 1, which is quite natural. However, be careful when learning to read/write code in other programming languages because some, e.g. Python, start counting at 0."

Since this PR contained multiple changes, it may be easier to just submit another PR to make these additional changes instead of reverting these commits. But I leave that decision to you.

01-supp-read-write-csv.Rmd
chendaniely: Probably meant to type `Read`?

dhaine:
Sorry for the delay to reply. I like very much the example given by @jdblischak . @emilliman5 , do you want to submit an other PR for this?

PR #120, opened 2015-05-25
Update 07-knitr-R.md

csoderberg:
added other output format types for knitR

### DIFF
--- a/07-knitr-R.md
+++ b/07-knitr-R.md
@@ -11,7 +11,7 @@ minutes: 30
 >
 > * To discover `knitr` and the generation of dynamic reports
 
-`Knitr` is an R package that makes your code neat, pretty, and shows your notes, code, and output simultaneously in an `html` document. You create these documents in `.Rmd` files. You can write in `LateX` or `md`. 
+`Knitr` is an R package that makes your code neat, pretty, and shows your notes, code, and output simultaneously in an `html`, `Word`, or `PDF` document. You create these documents in `.Rmd` files. You can write in `LateX` or `md`. 
 
 knitr extracts R code in the input document (.Rmd), evaluates it and writes the results to the output document (html). There are two types of R code: chunks (code as separate paragraphs) and inline R code.
 


dhaine:
Wouldn't it be more precise to say: you can write in LaTeX (`.Rnw` files) or Markdown (`.Rmd` files)?

jdblischak:
Thanks for the PR, @csoderberg.

For this repo, the source file to change is the Rmd version of the file. Both the md and html are automatically generated using the Makefile. Could you please update the Rmd version with your proposed change? See the contributing guidelines:

https://github.com/swcarpentry/r-novice-inflammation#contributing

As for the proposed change, I am a bit worried about the accuracy of the statement. From what I understand, knitr itself can convert Rmd -> md, Rnw -> pdf, and some other options ([ref](http://cran.r-project.org/web/packages/knitr/vignettes/knitr-refcard.pdf)). The knitr package also contains the convenience function `knit2html` which passes the md file it creates to the [markdown package](http://cran.r-project.org/web/packages/markdown/markdown.pdf) to create an html file. I believe this is the functionality that the original text was referring to.

Since this lesson was originally written, the [rmarkdown](http://rmarkdown.rstudio.com/) package was released. From what I understand, the rmarkdown package first converts Rmd -> md using knitr and then converts to html, pdf, or Word using pandoc.

So I guess my point is, if we want to discuss conversion of an Rmd file to pdf or Word, we should explicitly mention the rmarkdown package and explain how to use the `render` command. I would welcome a PR that updates this lesson accordingly.

Also, I agree with @dhaine that it should be more clear that Rmd files are written in Markdown. If you wanted to write in LaTeX, you'd have to write a Rnw file and convert it to pdf, not html (I'm not aware of a way to write in LaTeX and generate html). As it is written, it sounds as if you could write in LaTeX and convert it directly to html. If you don't have the time to update this lesson to include the rmarkdown package, I'd also be happy with just a small PR that makes this distinction clear.

jdblischak:
@csoderberg, what is the status of this PR? Please let us know if you need any assistance and/or guidance.

csoderberg:
The original lesson was pretty sparse, so it was difficult to determine exactly what it was referring to, but since no actual code was provided I assumed that students were being pointed to the 'knit to html' button at the top of the R script in RStudio, which has a drop down option for pdf and word, no additional commands required. 

I can move the PR over to the .rmd file, I figured it was just easier to reply to your comment here before moving it over.

dhaine:
Thanks for the follow-up @csoderberg . Looking back to this lesson, I believe the objective of the author was to demonstrate an easy way to generate reproducible reports, using markdown to generate an html document with the drop-down options of Rstudio. To mention LaTeX could confuse novice learners who don't know/don't want to know about it. Maybe you could re-write that first sentence of the lesson (in the Rmd) that we will show an easy way with markdown -> html (or Word) and mention LaTeX to be used with Rnw files only as a note.

jdblischak:
I'm closing this PR. Please feel free to reopen the PR with the Rmd file edited with our suggested changes. Please let me know if you need any help with this.

PR #123, opened 2015-06-06
Updating pointers to instructions and some instructions themselves

gvwilson:
Pointing at lesson-example instead of lesson-template, and telling people (again) not to commit HTML.


dhaine:
We prefer people to commit to the R Markdown file: please only commit changes to the Markdown files, *not* the generated HTML -> please only commit changes to the R Markdown files, *not* the generated Markdown/HTML

jdblischak:
@dhaine, I suggest we merge this PR and then you and I can do the work of reconciling CONTRIBUTING.md and https://github.com/swcarpentry/r-novice-inflammation#contributing. Since the lesson template is fairly stable now, I think we should modify CONTRIBUTING.md for this repo and then delete the directions in the README.

jdblischak:
@dhaine, this needs to be merged. What do you think of my suggestion? Please let me know if you want me to process this PR.

dhaine:
Sorry for the delay. Yes I can work on it if you can still wait another week...

jdblischak:
After #133, I am currently happy with the status of the contributing directions. I now think having the directions present in both the README and CONTRIBUTING is probably not a bad idea because it will increase the chances that a contributor will read it.

PR #126, opened 2015-06-11
small fixes on read csv supplementary lesson

jeremycg:
In the lesson on stripping white space, markdown is stripping the
whitespace (" Red" is being made as "Red") - added escape to fix.

Also added back in figures - they were removed (accidentally?). I renamed
them to fit the style.


jdblischak:
Thanks for the PR, @jeremycg.

Good catch on the missing figures. I had accidentally deleted them when I was removing unused figures from the repo.

Since you've already started improving the names of the figures, could you please fix them completely? The style we use is all lowercase with dashes for separation, e.g. `01-supp-csv-with-row-nums.png`.

jeremycg:
Thanks, I think that's it.

jdblischak:
Great. Thanks, @jeremycg!

PR #131, opened 2015-06-21
Added point 10 to best practices

kartucson:
Memory issues while running R code is common for computationally intensive scientific processes. Hence a simple rule-of-thumb is suggested to keep the 'memory buffer overflow' error under tab.

### DIFF
--- a/06-best-practices-R.Rmd
+++ b/06-best-practices-R.Rmd
@@ -69,13 +69,23 @@ rather than:
 dat <- read.csv(file = "/Users/Karthik/Documents/sannic-project/files/dataset-2013-01.csv", header = TRUE)
 ```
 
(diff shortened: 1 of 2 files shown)


jdblischak:
Thanks for the PR, @kartucson. Unlike the other lesson repos, we write the source in R Markdown and then render the Markdown and html from the Rmd files. Could you please make your changes to 06-best-practices-R.Rmd?

We're working on making the contributing guidelines for this repo easier to follow, but for now see
https://github.com/swcarpentry/r-novice-inflammation#contributing

jdblischak:
Thanks, @kartucson. Sorry it took me so long to merge. I only get an email update when a new message is posted, not when a new commit is added.

PR #134, opened 2015-06-26
Update 01-supp-intro-rstudio.Rmd

willpitchers:
I added a tip callout about tab-completion to the RStudio introduction.

### DIFF
--- a/01-supp-intro-rstudio.Rmd
+++ b/01-supp-intro-rstudio.Rmd
@@ -39,6 +39,12 @@ on if needed. You can copy-paste into the R console, but the Rstudio script
 editor allows you to 'send' the current line or the currently selected text to
 the R console using the `Ctrl-Enter` shortcut.
 
+> ## Tip {.callout}
+>
+> Rstudio is pretty smart when it comes to recognizing commands.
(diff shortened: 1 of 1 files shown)


jdblischak:
Thanks for the PR, @willpitchers.

I like the content, but what about moving it out of the callout box? Unlike the other lessons, this entire lesson is little tidbits. Callout boxes are for when you don't want to disrupt the main narrative. I don't think tab-completion needs to be treated differently than switching between windows or sending code from a script to the console. What do you think?

willpitchers:
Hey John – that's a good point. I'm fine with moving it out of the callout
box.

W

jdblischak:
OK. Great. Please respond to the thread once you make the change to remind me to merge.

willpitchers:
I've made the edit in my forked version – I guess that it'll appear as a separate PR now?
Cheers,  W

jdblischak:
I'm not quite sure what happened because this PR says it is from an "unknown repository". Normally you could just update the branch in your fork and the PR would be automatically updated.

In this case, you'll need to create a new PR from your branch `willpitchers-patch-1`.

PR #135, opened 2015-07-02
Add common gotchas with git in RStudio to instructor notes

karawoo:
Two frequent issues with using git in RStudio are making sure RStudio knows where to find git and making sure to set an upstream reference so that RStudio can push and pull directly to/from GitHub.

### DIFF
--- a/instructors.Rmd
+++ b/instructors.Rmd
@@ -307,9 +307,11 @@ cat line-count.R
 
 Some instructors will demo RStudio's git integration at some point during the
 workshop. This often goes over very well, but there can be a few snags with the
-setup. First, RStudio may not know where to find git. You can specify where git
-is located in _Tools > Global Options > Git/SVN_; on Mac/Linux git is often in
-`usr/bin/git` or `usr/local/bin/git` and on Windows it is often in
+setup. First, RStudio may not know where to find git (this shouldn't be a
+problem for Windows users who installed Git Bash and selected the option to run
+Git from the Windows command prompt). You can specify where git is located in
+_Tools > Global Options > Git/SVN_; on Mac/Linux git is often in `usr/bin/git`
+or `usr/local/bin/git` and on Windows it is often in
 `C:/Program Files (x86)/Git/bin/git.exe`. If you don't know where git is
 installed on someone's computer, open a terminal and try `which git` on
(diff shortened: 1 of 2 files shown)


sritchie73:
Nice!

jdblischak:
Thanks for the PR, @karawoo.

The only thing I can think to add is that when Windows users are installing Git Bash, if they choose the option to run Git from the Windows command prompt, then RStudio automatically finds the Git executable.

Please let me know if you plan to add on this information or would prefer to leave it as is. I'm happy to merge the PR either way.

karawoo:
Oh interesting @jdblischak, I didn't know that. I'll add this information.

karawoo:
Ok I added that note, let me know if you'd rather these commits be squashed together instead of separate.

jdblischak:
Thanks! And no need to squash the commits. The commit history of this repo is a beautiful mess at this point, especially with the separate commits to render the html.

Though I must apologize that I did a poor job explaining the Windows situation. In order for that solution to work, Windows users would need to manually choose the non-default option "Run Git from the Windows Command Prompt" during the installation process ([more details](http://guides.beanstalkapp.com/version-control/git-on-windows.html#installing-git)). This direction is not currently in the main workshop [setup directions](http://swcarpentry.github.io/workshop-template/#setup) because it is specific only to those workshops where the RStudio Git GUI is being used. Thus I'd recommend saying something to the effect of "If you're planning on using Git via RStudio in your workshop, you should edit the workshop setup instructions for Git Bash to include the step..."

karawoo:
Ah ok, I will rephrase. I've never had people select the option to run Git Bash from the Windows command prompt before -- does that mean that the bash commands from the shell lesson are also available in the command prompt, or do they still need to use the Git Bash shell for that?

jdblischak:
It is my understanding that choosing this option allows one to use Git commands in the Windows shell, but other than that the shell will behave like a normal Windows shell. In other words, I don't think shell commands will work in the Windows shell.

The only reason to choose this option is so that RStudio automatically detects the Git executable. You would still use Git Bash for running shell commands or running Git at the command line.

karawoo:
Thanks @jdblischak, hopefully I've made it clearer in this update.

jdblischak:
Great. Thanks!

PR #140, opened 2015-07-15
Changed text to combine 2 sentences.

jessicaguo:


dhaine:
Thanks @jessicaguo. Maybe you could also add a note that the indentation is automatically inserted by Rstudio (and you could -or shouldn't- modify it in the Global Options menu)?

jdblischak:
I like this change. Since not everyone uses RStudio, I don't think it necessary to provide specific instructions for it. Thanks for the PR, @jessicaguo.

PR #143, opened 2015-08-18
added introduction text+ inflammation data downlaod

areejalsheikh:
I figured at workshops when students only take R, they need a separate intro even if it's a replicate of the one at Python. More importantly, having a dedicated link to R-inflammation data.

P.S: current link to the R-inflammation data is broken. Could you please fix that?

### DIFF
--- a/index.md
+++ b/index.md
@@ -53,9 +53,8 @@ To do all that, we'll have to learn a little bit about programming.
 > ## Getting ready {.getready}
 >
 > You need to download some files to follow this lesson:
-> NOTE: If you have done the Python lesson, then you should have the same inflammation data. If so, ignore this step.
 >
-> 1. Make a new folder in your Desktop called `R-novice-inflammation`.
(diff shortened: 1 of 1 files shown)


dhaine:
Thanks for the PR @areejalsheikh ! This would make our r-novice-inflammation index in line with the one for python. Could you just change R-novice-inflammation to r-novice-inflammation? I know it's capital R but we use small r everywhere to name this lesson...
There is no zipped file for the data sets and I'll add that. Note that we have a few extra datasets than the python lesson, so it's better removing your note on ignoring this step.

areejalsheikh:
Done but I'm not sure if you can see it. Do I need a pull request for that too?

dhaine:
No need for a PR, the changes are there. Just one small change though. Could you change R-novice-inflammation-data.zip to r-novice-inflammation-data.zip? Then I'll be able to accept your changes. Thanks!

jdblischak:
Thanks for the PR, @areejalsheikh. It was recently incorporated in #152.

PR #145, opened 2015-08-25
Full stops added to objectives.

raiphilibert:


dhaine:
Thanks @raiphilibert for this PR. Maybe someone can argue on the use of full stops, comma or semi-colon in bullet list. To me what you use should be constant across the document (want to check the other modules of this lesson?). Apparently people decided to use full stops in the other lessons (shell, python). So for now I'm accepting your PR.

PR #147, opened 2015-10-06
Change misleading wording about return statements

mxtthias:
Implicit returns are not unique to R. New wording is more general.


mxtthias:
Closes #147

jdblischak:
Thanks, @mxtthias!

PR #155, opened 2015-12-08
Use relative path in relative path example

botanize:
The path shown in 06-best-practices-R.Rmd ```/files/dataset-2013-01.csv``` was an absolute path, not a relative path.


jdblischak:
Good catch! Thanks, @botanize.

PR #156, opened 2016-01-04
added Sarah Meister's challenge

thesarahmeister:
Created a challenge where the student completes a partially done function.

### DIFF
--- a/02-func-R.Rmd
+++ b/02-func-R.Rmd
@@ -532,3 +532,35 @@ max(answer)
 >
 > but the chances of us typing all 12 filenames correctly aren't great, and we'll be even worse off if we get another hundred files.
 > What we need is a way to tell R to do something once for each file, and that will be the subject of the next lesson.
+
+
+> ## Sarah's Challenge - Another function {.challenge}
+>
+>  Below is an incomplete function that calculates some basic statistical descriptions of a vector, `x`. Add to the function so that it automatically creates a dataframe output that includes each of the variables calculated by the function. Be sure to give columns of the dataframe descriptive names. Use the `round` function to round each result of the output to 3 places.
+>  Be sure to document each step of the function with comments.
+
+```{r rescale-test, include=TRUE}
+describe = function(x)
++ {
++    m <- mean(x,na.rm=T)
++    s <- sd(x,na.rm=T)
++    N <- sum(is.na(x))
++    n <- length(x)-N
++    se <- s/sqrt(n)
+}
+```
+
+```{r rescale-test, include=FALSE}
+describe = function(x)
++ {
++    m=mean(x,na.rm=T)
++    s=sd(x,na.rm=T)
++    N=sum(is.na(x))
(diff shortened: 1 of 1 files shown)


02-func-R.Rmd
```diff
@@ -532,3 +532,35 @@ max(answer)
 >
 > but the chances of us typing all 12 filenames correctly aren't great, and we'll be even worse off if we get another hundred files.
 > What we need is a way to tell R to do something once for each file, and that will be the subject of the next lesson.
+
+
+> ## Sarah's Challenge - Another function {.challenge}
```
jdblischak: Please remove your name from the title of the challenge.

02-func-R.Rmd
```diff
@@ -532,3 +532,35 @@ max(answer)
...
+>
+>  Below is an incomplete function that calculates some basic statistical descriptions of a vector, `x`. Add to the function so that it automatically creates a dataframe output that includes each of the variables calculated by the function. Be sure to give columns of the dataframe descriptive names. Use the `round` function to round each result of the output to 3 places.
+>  Be sure to document each step of the function with comments.
+
+```{r rescale-test, include=TRUE}
```
jdblischak: Chunks in the same file must each have a unique name. There is already a chunk named "rescale-test" in this file. Also, the `include=TRUE` is unnecessary because it is the default.

02-func-R.Rmd
```diff
@@ -532,3 +532,35 @@ max(answer)
...
+describe = function(x)
++ {
++    m <- mean(x,na.rm=T)
++    s <- sd(x,na.rm=T)
```
jdblischak: Since this is part of the challenge, each line needs to start with a `>`. See the other challenges in this lesson for examples.

jdblischak:
Thanks for the PR, @thesarahmeister. I've added some inline comments. Here are some other things to work on:

*  The code style. This is an open source projects with lots of contributors. To make it easy for a learner to read the lesson, we need to use a uniform coding style. Please look at the other code in this lesson to learn the style. When in doubt, use the recommendation in [Hadley's style guide](http://adv-r.had.co.nz/Style.html).
*  What is the exact thing you are testing with this challenge? Your challenge does require the learner to remember to use `return`, which is good. But the other lines of code they need to write are not related to writing functions. In r-novice-inflammation, we do not cover the idea that vectors can be named nor the function `round`. Thus this challenge would be very difficult for learners to complete.

Please let me know if you have any questions.

thesarahmeister:
okay great so do I need to re-do the challenge then? As part of training?

gvwilson:
No need to re-do to move onto the next step, but if you'd like to, we'd 
welcome it.

jdblischak:
This PR still has lots of formatting issues that need to be addressed, and the challenge mainly tests skills not addressed in the lesson. I'm going to close it. Please feel free to update and re-open. And of course please ask if you'd like any help.

PR #158, opened 2016-01-04
Remove call stack section

c-martinez:
The Call Stack section of [the python counterpart of this lesson](https://swcarpentry.github.io/python-novice-inflammation/06-func.html) has been removed from the lesson and moved to the [discussion page](https://swcarpentry.github.io/python-novice-inflammation/discussion.html).

As mentioned [in this PR](https://github.com/swcarpentry/python-novice-inflammation/pull/178), it would be good to know whether instructors teaching this lesson tend to skip this section and thus it could be removed?

My guess is that, ideally, we would like the R and Python lessons to be as similar as possible (except when it comes to language-specific lessons). Is this right?

### DIFF
--- a/02-supp-call-stack.Rmd
+++ b/02-supp-call-stack.Rmd
@@ -1,0 +1,106 @@
+### The Call Stack
+
+Let's take a closer look at what happens when we call `fahr_to_celsius(32)`. To make things clearer, we'll start by putting the initial value 32 in a variable and store the final result in one as well:
+
+```{r}
+original <- 32
+final <- fahr_to_celsius(original)
+```
+
+The diagram below shows what memory looks like after the first line has been executed:
+
+<img src="fig/python-call-stack-01.svg" alt="Call Stack (Initial State)" />
+
+When we call `fahr_to_celsius`, R *doesn't* create the variable `temp` right away.
+Instead, it creates something called a [stack frame](reference.html#stack-frame) to keep track of the variables defined by `fahr_to_kelvin`.
+Initially, this stack frame only holds the value of `temp`:
+
+<img src="fig/python-call-stack-02.svg" alt="Call Stack Immediately After First Function Call" />
+
+When we call `fahr_to_kelvin` inside `fahr_to_celsius`, R creates another stack frame to hold `fahr_to_kelvin`'s variables:
+
+<img src="fig/python-call-stack-03.svg" alt="Call Stack During First Nested Function Call" />
+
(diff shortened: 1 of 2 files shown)


jdblischak:
Thanks for the PR, @c-martinez. I agree we should remove this section from the main lesson. While instructive for reading, it is difficult to teach in a workshop setting. In r-novice inflammation, we often put supplemental content in "supp" files. Could you please put the call stack section in a new section titled `02-supp-call-stack.Rmd`? Then add a callout section in `02-func-R.Rmd` that links to this supplemental content. Please let me know if you have any questions.

c-martinez:
Hi @jdblischak , I've added the `02-supp-call-stack.Rmd` fine and a `.callout` section. I hope this is fine now.

jdblischak:
It's looking good, @c-martinez. Three minor things.

Each file needs a YAML header so that it is properly rendered online. Please add the following to the very top of the file `02-supp-call-stack.Rmd`.

```
---
layout: page
title: Programming with R
subtitle: The call stack
minutes: 15
---
```

The callout section as it is has no context. The word "call stack" is not used before this point, so the title "Further details" is not correct since no details on the call stack have been given. Could you please title the callout box "The call stack" and then give a brief description? Maybe something like "For a deeper understanding of how functions work, you'll need to learn how they create their own environments and call other functions. Function calls are managed via the call stack. For more details on the call stack, have a look at the [supplementary material](02-supp-call-stack.html)."

The call stack is still included in the key points at the bottom of the lesson. Could you please move this key point to the supplementary file?

c-martinez:
Hi @jdblischak, I've incorporated the changes you suggested. I hope this is fine now.

02-supp-call-stack.Rmd
```diff
@@ -0,0 +1,113 @@
...
+The big idea here is [encapsulation](reference.html#encapsulation), and it's the key to writing correct, comprehensible programs.
+A function's job is to turn several operations into one so that we can think about a single function call instead of a dozen or a hundred statements each time we want to do something.
+That only works if functions don't interfere with each other; if they do, we have to pay attention to the details once again, which quickly overloads our short-term memory.
+
+> ## Challenge - Following the call stack {.challenge}
+>
+>  + We previously wrote functions called `fence` and `outside`.
+>    Draw a diagram showing how the call stack changes when we run the
+>    following:
+> ```{r, results="hide"}
```
chendaniely: from the `knitr` documentation: `hide` hides results; this option only applies to normal R output (not warnings, messages or errors)

if we actually execute the below block of code in this document, we have an error: 
`## Error in eval(expr, envir, enclos): could not find function "outside"`

maybe use `eval=FALSE` instead?

PR #161, opened 2016-01-07
Exercise added to the lesson Making Choices (Conditionals)

AntoniaGogoglou:


jdblischak:
Thanks for the PR, @AntoniaGogoglou.

I don't think novice learners are going to come up with the answer to your challenge. They are totally unfamiliar with using these file I/O functions, so they aren't going to understand what the problem is.

Also, what is the use case for the code as you have written it? I realize it produces the bug you want, but otherwise it seems unnecessarily complicated.

```
# original
WriteToFile<-function(InputVector, out = NULL){
  newInputVector<-InputVector-min(InputVector)
  f1<-file(out)
  write(newInputVector,f1)
  close(f1)
}
# easier with write
WriteToFile<-function(InputVector,out = stdout){
  newInputVector<-InputVector-min(InputVector)
  write(newInputVector, file = out)
}
# even easier with cat
WriteToFile<-function(InputVector, out = ""){
  newInputVector<-InputVector-min(InputVector)
  cat(newInputVector, file = out)
}
```

How about this for an idea? Give them this basic function, but then have them add an `if` to only output the vector if it's length is greater than, say, 5.

If you plan to update the PR, please also address the following formatting issues:
*  Submit edits to the Rmd, not the md file
*  Follow the code style used throughout the lesson
*  Use backticks when referring to R functions, not quotes
*  Include the answer in a code chunk with the options `eval=FALSE, include=FALSE`

Please let me know if you have any questions.

PR #164, opened 2016-01-11
added exercise for instructor training requirement

coleoguy:
This exercise asks the student to make a slide show and explore a couple of the central components of most slide decks.

### DIFF
--- a/07-knitr-R.Rmd
+++ b/07-knitr-R.Rmd
@@ -52,3 +52,15 @@ A code chunk will set off the code and its results in the output document, but y
 > 2. Copy code from earlier into code chunks to read the inflamation data and plot average inflammation.
 > 3. Add a few notes describing what the code does and what the main findings are. Include an in-line calculation of the median inflamation level.
 > 4. `knit` the document and view the html result.
+
+
+
+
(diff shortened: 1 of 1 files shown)


jdblischak:
Thanks for the PR, @coleoguy.

This challenge is very ambitious. Could you please provide some more explanation for the learners? For example, you say to create an equation, but most won't even know where to start. Maybe give them an equation to use? Also, you say to use `$$`, but the example equation used earlier in the lesson used `$`. You'll need to explain why they should use `$$` in this case instead of `$`.

But before that, you need explain some how this file works, e.g. that slides are demarcated by level two headers. Also a transition would be useful. They just learned how to make html reports, and all of a sudden now they are making presentations. Most will be familiar with programs like PowerPoint and Keynote, and will not associate HTML with the ability to make presentations. You'll need to provide this background context.

Also, please make your changes to the Rmd file, not the md.

https://github.com/swcarpentry/r-novice-inflammation/blob/gh-pages/CONTRIBUTING.md

gvwilson:
I think this is useful, but probably out of reach for our novices.

PR #165, opened 2016-01-21
Exercise for instructor training

gharanczyk:


jdblischak:
Thanks for the PR, @gharanczyk.

At this point in the lesson, the idea that functions can take multiple named arguments has not been explicitly addressed. This comes in the next lesson. In this lesson, they've only briefly seen named arguments with `read.table` and multiple unnamed arguments with `apply`. But while they've seen those examples, this is not something they can apply yet. One idea to make it easier would be to add an example first. For example, add a title with `main`, and then task them with labeling the x and y axes.

PR #166, opened 2016-01-21
add challenge for read.csv with header = TRUE

lmweber:
This challenge deepens understanding of the `read.csv` function and data frames.


chendaniely:
@lmweber I had to change the

> `read.csv` with `header = TRUE`

bits of your code, since it get's parsed incorretly:

screenshot from 2016-06-14 13-07-24

if you don't mind, I've changed the title to "Loading data with headers"

If you'd like you can submit a PR to change the name

PR #168, opened 2016-02-02
Additional exercise for "Using loops" challenge

tinyendian:
Here is an additional exercise as requested in the SWC instructor training. The proposed exercise builds on the contents of the preceding exercises (the seq function, computing a total, using exponentiation) and applies them to solve the "Wheat and Chessboard" problem. The solution can be compared with a mathematical solution.

The last question in the exercise technically introduces the concept of numerical precision, which may be undesirable.

Variants of the exercise could, e.g., ask learners to reuse the number of grains from a previous loop iteration instead of applying the suggested mathematical expression, although this could make the problem a bit harder to solve.

### DIFF
--- a/03-loops-R.Rmd
+++ b/03-loops-R.Rmd
@@ -214,6 +214,20 @@ length(vowels)
 > expo(2, 4)
 > ```
 >
+> 4. Solving the "Wheat and Chessboard" problem
+>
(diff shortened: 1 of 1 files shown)


chendaniely:
Hi @tinyendian thanks for the PR.

I agree that the introducing numerical precision might be out of scope with what we are trying to teach in this lesson.  If you'd like you can re-submit the PR with only the total grain count portion.

Also, can you also have add another code block that has the solution?
We are trying to get better at providing solutions for our instructors.

you can use the knitr `eval=FALSE and echo=FALSE` to hide the solution.

For the SWC instructor training, you've done the task correctly. :+1: 

I am going to close this PR for the Version 5.4 milestone, please feel free to resubmit the PR.

tinyendian:
Hi @chendaniely,

Thanks for your comment. I don't feel strongly about this - I can resubmit it with the suggested changes if there is interest in having this additional exercise in the lesson, but I'm also happy to discard it if there are already enough exercises.

PR #176, opened 2016-02-12
Added exercise for instructor training requirement

LouisRanjard:


04-cond.Rmd
```diff
@@ -320,6 +320,31 @@ analyze <- function(filename, output = NULL) {
 }
 ```
 
+> ## Challenge -  {.challenge}
```
gvwilson: Needs a better title.

gvwilson:
Thanks for the PR - can you please modify to include a meaningful title? And @hdashnow or @chendaniely can you please check the content?

hdashnow:
Thanks for the challenge @LouisRanjard 
I think this challenge would fit better slightly earlier in the lesson, for example just before or just after the challenge "Using conditions to change behaviour".
It also needs a title, and a solution would be nice, too.
What do you think @LouisRanjard? Are you happy to make some changes?

04-cond.Rmd
LouisRanjard: What about: 
Challenge - Find the maximum inflammation score
?

04-cond.Rmd
```diff
@@ -320,6 +320,31 @@ analyze <- function(filename, output = NULL) {
...
+```{r, echo=FALSE}
+filenames <- list.files(path = "data", pattern = "csv", full.names = TRUE)
+filename_max <- "" # filename where the maximum average inflammation patient is found
+patient_max <- 0 # index (row number) for this patient in this file
+average_inf_max <- 0 # value of the average inflammation score for this patient
+for (f in filenames) {
+  dat <- read.csv(file = f, header = FALSE)
+  rownum <- 1
+  for (patient_average_inf in apply(dat, 2, mean)){
+    # Add code here ...
```
LouisRanjard: Solution:
if (patient_average_inf>average_inf_max) {
  filename_max <- f
  patient_max <- rownum
}

LouisRanjard:
Hi there @hdashnow , yes I agree it would probably fit best after the challenge "Using conditions to change behaviour". I have added a title "Find the maximum inflammation score" and the solution in line notes (but for some reason I cannot edit my pull request).

Issue #129, opened 2015-06-18
Indosyncracy of mean() on inflammation-01


erex:
This works as expected
```
# mean inflammation on day 7
mean(dat[, 7])
```
but this fails
```
# mean of measurements for patient 7
mean(dat[7,])
```
returns an error.  That is because there is no header in this file and arbitrary field names are give.

Might be worthwhile to note that it makes no medical sense to calculate the mean of 24 measurements.

jdblischak:
Thanks for reviewing our lesson, @erex.

The issue you have found is an unfortunate artifact of how this lesson was developed. It is a translation of [novice-python-inflammation](https://github.com/swcarpentry/python-novice-inflammation).

When you slice a single column of a data frame, it returns a vector. When you slice a single row, it returns a data frame. The function `mean` does not work on a data frame (because a data frame often has data of different types). You can observe this behavior by running the following:

```r
class(mean(dat[, 7]))
class(mean(dat[7, ]))
```

The more natural way to store a bunch of numbers would be in R's `matrix` class. Then one could take the mean of a row. However, it was decided during the translation of this lesson that it would be better to use a data frame since it is so ubiquitous in R programming that novices should be exposed to it.

We have documented this [here](https://github.com/swcarpentry/r-novice-inflammation/blob/gh-pages/discussion.md#error-returned-when-attempting-to-calculate-mean-of-dataframe-rows) so that our instructors know about this quirk when teaching a workshop.

If you'd like to add a callout box to warn learners about this behavior, I'd welcome a Pull Request.

> Might be worthwhile to note that it makes no medical sense to calculate the mean of 24 measurements.

I feel that is a very strong opinion for such a context-dependent decision. The usefulness of a summary statistic depends on both the statistical properties of the data and the question being addressed.

jdblischak:
@erex, are you interested in creating a callout box that warns learners about taking the mean of a row of a data frame?

erex:
@jdblischak thanks for the invitation; I've got a fair bit on my plate for the next few days.  I'll try to do as you ask next week.  If I've not sent a pull request to you by the end of next week, I've gotten overtaken by events.

dhaine:
Maybe with a mention about rowMeans()?

lmweber:
[...]

PR #177, opened 2016-02-12
use as.numeric() to convert row of data.frame to vector

lmweber:
On some installations of R, running commands such as max / min / mean on a sliced row of a data.frame returns an error, since R does not automatically convert the row of data to a numeric vector (confusingly, columns are automatically converted). During a recent workshop we solved this by using the as.numeric() command. I have added a Tip explaining this issue and suggesting the use of as.numeric() if others encounter the same problem.

### DIFF
--- a/01-starting-with-data.Rmd
+++ b/01-starting-with-data.Rmd
@@ -240,6 +240,19 @@ patient_1 <- dat[1, ]
 max(patient_1)
 ```
 
+> ## Tip {.callout}
+>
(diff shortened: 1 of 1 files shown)


lmweber:
This pull request also addresses the following issue: https://github.com/swcarpentry/r-novice-inflammation/issues/129

chendaniely:
thanks @lmweber I think it's good to at least have a mention of why it works in one margin and not the other.  Seems like a good compromise between everything mentioned in #129

PR #159, opened 2016-01-06
Exercise for instructor training

kyrcha:
Add new challenge for finding the max of columns

PR #179, opened 2016-02-16
added another slicing exercise

gbass:
I created this exercise to fulfill the requirements of the instructor training course.  I also added the solution below the exercise.

### DIFF
--- a/01-starting-with-data.Rmd
+++ b/01-starting-with-data.Rmd
@@ -330,6 +330,25 @@ We'll learn why this is so in the next lesson.
 > (c) `max(dat[5, 3:7])`
 > (d) `max(dat[5, 3, 7])`
 
+
+> ## Challenge - Slicing and re-assignment {.challenge}
+> 
+> Let's pretend there was something wrong with the instrument on the first five days for every second patient (#2, 4, 6, etc.), which resulted in the measurements being twice as large as they should be.
+>
+> 1. Write a vector containing each affected patient (hint: `? seq`)
+> 2. Halve the first five days' values in only those patients
+> 3. Print out the data frame to check that your code has fixed the problem
+>
+>> Solution:
+>> ```{r}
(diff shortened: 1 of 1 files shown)


01-starting-with-data.Rmd
```diff
@@ -330,6 +330,25 @@ We'll learn why this is so in the next lesson.
 > (c) `max(dat[5, 3:7])`
 > (d) `max(dat[5, 3, 7])`
 
+
+> ## Challenge - Slicing and re-assignment {.challenge}
+> 
+> Let's pretend there was something wrong with the instrument on the first five days for every second patient (#2, 4, 6, etc.), which resulted in the measurements being twice as large as they should be.
```
chencindyj: I think you should mention something like "To follow up from the previous data frame" to ensure that they know the question is related to the previous one.  Otherwise, they might make the vector as instructed in the first point, and they might get confused that they were supposed to turn it into a data frame afterwards.

gvwilson:
@chendaniely merge?

chendaniely:
@gvwilson @hdashnow was there a standard way solutions should be provided?

I've also seen hiding the solution so it does not render on the page, but still available in the source

hdashnow:
@chendaniely here's the suggested solution format (from #159 )

@gbass could you please format the answer like this so it doesn't appear in the notes:
```{r slicing-re-assignment-answer, eval=FALSE, include=FALSE}
whichPatients <- seq(2,40,2)
whichDays <- c(1:5)
dat[whichPatients,whichDays] <- dat[whichPatients,whichDays]/2
(dat)
```

hdashnow:
Hey @gbass, I've made the suggested edits and submitted them as a new PR #209.
I've made a few changes, so please check out #209 and let me know if you'd like to change anything. Otherwise, I'll merge it tomorrow.
Thanks for contributing!

Issue #183, opened 2016-02-25
Typo in Lesson 1 setwd function call


pschloss:
The lesson currently has:

    setwd("~/Desktop/r-novice-inflammation/)

It should be...

    setwd("~/Desktop/r-novice-inflammation/")

pschloss:
@chendaniely I *think* I filed a PR for this back in February, can I get it merged?

chendaniely:
@pschloss merging!

hdashnow:
Thanks! Fixed by merging #184

PR #184, opened 2016-02-25
Added quote to end of setwd command

pschloss:
Addresses issue #183


pschloss:
This should close the issue

gallington:
This fix appears to be correct.

PR #185, opened 2016-02-28
added subsetting tip for instructor training

jacel:
Although the inflammation file doesn't have a header, it's often really useful to address columns by name. During an R lesson I taught, this was one of the first questions I had from a student. Since there's already a supplementary lesson that goes into this in detail, I think it's worth mentioning during the main lesson. Cheers!


gvwilson:
Looks OK to me - merge?

chendaniely:
fixed in #199

PR #186, opened 2016-03-01
r-novice-inflammation modified in 04-cond files

brugerer:
Added comment about 'else if' statements in 04-cond.* files.  Specific text added was:
"Note that when combining `else` and `if` in an `else if` statement (similar to `elif` in Python), the `if` portion still requires a direct input condition.  This is never the case for the `else` statement alone, which is only executed if all other conditions go unsatisfied."

### DIFF
--- a/04-cond.Rmd
+++ b/04-cond.Rmd
@@ -143,7 +143,7 @@ sign(0)
 sign(2/3)
 ```
 
-Note that when combining 'else' and 'if' in an 'else if' statement (similar to 'elif' in Python), the 'if' portion still requires a direct input condition.  This is never the case for the 'else' statement alone, which is only executed if all other conditions go unsatisfied. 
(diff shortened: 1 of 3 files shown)


gvwilson:
Thanks for the PR - can you please re-do with only the .Rmd file?  We'll regenerate the .md and .html after merge.  @hdashnow and @chendaniely, can you please decide if you'd like the content?

04-cond.Rmd
```diff
@@ -143,6 +143,7 @@ sign(0)
 sign(2/3)
 ```
 
+Note that when combining `else` and `if` in an `else if` statement (similar to `elif` in Python), the `if` portion still requires a direct input condition.  This is never the case for the `else` statement alone, which is only executed if all other conditions go unsatisfied.
```
chendaniely: do not need reference to Python.

hdashnow:
It just so happened that the html merge worked!

PR #188, opened 2016-04-12
Instructor-training: added challenge for 02-func-R on scoping rules

davfre:
NB! Submitted as part of the requirements for the instructor training

The purpose of this multiple choice question is to test the understanding of R's:

- lexical scoping (primarily)
- use of named function arguments
- default argument values (and overwriting them)

### DIFF
--- a/02-func-R.Rmd
+++ b/02-func-R.Rmd
@@ -232,6 +232,29 @@ That only works if functions don't interfere with each other; if they do, we hav
 > result <- outside(fence(inner_vec, outer_vec))
 > ```
 
+> ## Challenge - named variables and the scope of variables {.challenge}
+>
+>  + Functions can accept arguments explicitly assigned to a variable name in
+>    in the function call `functionName(variable = value)`, as well as arguments by
+>    order:
+> ```{r}
+> a = 20
+> mySum = function(a, b = 10) {
+>   a = a + b
+>   return(a)
+> }
+> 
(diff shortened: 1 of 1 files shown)


tomwright01:
As you point out in your comment, this exercise is introducing two new concepts: *named vs positional arguments* and *default values*. Perhaps this could be extended with a more open ended question - 'Why does `mySum(3) == 13`, why does `mySum(b=3)` return an error?

A couple of minor comments on style, could you use the assignment operator `<-` rather than `=`. Also could you use more descriptive names for the arguments (e.g. input_1, input_2) instead of a & b?

davfre:
Great suggestions, thank you! I made the style changes and added your open ended question.

02-func-R.Rmd
```diff
@@ -232,6 +232,27 @@ That only works if functions don't interfere with each other; if they do, we hav
 > result <- outside(fence(inner_vec, outer_vec))
 > ```
 
+> ## Challenge - named variables and the scope of variables {.challenge}
```
gvwilson: We can remove the word 'Challenge' (the formatting will handle that).

gvwilson:
Looks good to me.

chendaniely:
@davfre can you reformat the question such that it doesn't use a multiple choice answer?
This way the question format is consistent with the rest of the lesson.

Issue #169, opened 2016-02-07
Not best practice in the final inflammation example


arohl:
In lesson 4, when the students get their R program to the point where they have a single command to generate all the plots, there is a bit of a problem. If they run the script a second time, they open the newly created pdf documents as well as the csv files and the csv reader gets a little upset being given a pdf file ;-)

In the example before this, where they run a command to generate the plot, the plot is put in the top level directory of the exercise and the data is in the data subdirectory. When they go to automating this, the example reads the cvs file from the data subdirectory and then also writes the pdf to the same directory. My fix (and I am no R programmer!) was to get the automatic program to read the csv filenames **without** the path, do the substitution to create the pdf filename and then prepend *data/* to the csv filename, which results in the PDF files all being in the working directory, whilst reading the cvs from the data subdirectory. Another alternative would be to make sure that only filenames with *inflammation* and *csv* are opened. I leave it to the experts to decide which is preferable.

  Andrew

As an aside, a very nasty result of the R program is that if in the string replacement, the student mistypes csv (e.g. as cvs), all of their input files are clobbered as the pdf filename is the same as the csv filename i.e. they write to the input file - the first solution above will also fix that!

hdashnow:
Is this fully resolved by #189 and #212? Shall I close this issue?

arohl:
Yes, please close

hdashnow:
Ta

PR #189, opened 2016-05-09
Edit analyze_all function in lesson 04-cond.Rmd

jrnold:
The analyze_all function would save files in the same directory as the data csv files.
This would cause problems on repeated runs without updating the pattern used to select files. Since using a pattern to only get csv files matching inflation.*\\.csv would require mentioning regular expressions, I opted to fix this by saving the figures to a different directory.

I also used this as an opportunity to reference the good scienctific computing practice of separating data and output referenced in "Good Enough Practices for Scientific Computing."

This fixes issue #169.

### DIFF
--- a/04-cond.Rmd
+++ b/04-cond.Rmd
@@ -251,36 +251,61 @@ is.null(output)
 !is.null(output)
 ```
 
-Now we can use `analyze` both interactively:
+Now we can use `analyze` interactively, as before,
 
 ```{r inflammation-01}
 analyze("data/inflammation-01.csv")
 ```
 
-and to save plots:
+but also use it to save plots,
 
 ```{r results='hide'}
 analyze("data/inflammation-01.csv", output = "inflammation-01.pdf")
 ```
 
-This now works well when we want to process one data file at a time, but how can we specify the output file in `analyze_all`?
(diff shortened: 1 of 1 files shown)


arokem:
This is an *instructor training* PR (I can't label it).

chendaniely:
@jainsley sidenote: we typically teach `bash` before `R`, where we show students the `?` and `*` wildcards for globbing.  You don't necessarily *have* to go down the regex route to teach pattern matching.

jrnold:
that's true, but all the R functions that take pattern arguments use regexes, which in this case would look something like `"inflammation.*\\.csv"`. We could use `glob2rx("inflammation*.csv")`, but then we'd still need to explain, or at least mention, what regexes are because we need to explain why we are using `glob2rx()`.

gvwilson:
Thanks for the PR @jrnold - @chendaniely, shall we take as-is or do you think the pattern-matching will be a step too far for our learners?

jrnold:
I don't think it would be bad to introduce pattern-matching. However, since the purpose of doing so in this case would be to allow them to use an anti-pattern, I don't think it should be used here.

hdashnow:
Thanks @jrnold. I rather like this as-is. If we you/we decide to go down the pattern-matching path later just pop in another PR

PR #223, opened 2016-07-08
Merging new styles and documentation

gvwilson:
This PR brings in all of the 2016-06 template, along with documentation on building lessons.

PR #225, opened 2016-07-09
Fix path for figures + Fix #224

### LINKED ISSUE #224: Fix errors in new-format episodes
Several of the episodes in #223 have errors due to @gvwilson's lack of R skillz. Look for `.error` in `_episodes/*.md` to find them, then trace back to the `_episodes_rmd/*.Rmd` file to fix.

fmichonneau:
The regular expression used to detect the CSV files in the data folder relied on the presence of the string "inflammation". Now, it looks like there are PDF files that also match the regular expression used so the script is trying to read in these PDF files causing the error messages. 
 I adjusted the regular expression accordingly (but it might actually be best to remove these PDF files?).

### DIFF
--- a/_episodes_rmd/01-starting-with-data.Rmd
+++ b/_episodes_rmd/01-starting-with-data.Rmd
@@ -25,7 +25,7 @@ keypoints:
 
 ```{r, include = FALSE}
 source("../bin/chunk-options.R")
-opts_chunk$set(fig.path = "fig/01-starting-with-data-")
+knitr_fig_path("01-starting-with-data-")
 ```
 
 We are studying inflammation in patients who have been given a new treatment for arthritis,
--- a/_episodes_rmd/02-func-R.Rmd
+++ b/_episodes_rmd/02-func-R.Rmd
@@ -23,7 +23,7 @@ keypoints:
 
 ```{r, include = FALSE}
 source("../bin/chunk-options.R")
-opts_chunk$set(fig.path = "fig/02-func-R-")
+knitr_fig_path("02-func-R-")
 ```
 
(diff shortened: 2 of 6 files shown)


gvwilson:
04-cond.Rmd has `pdf("inflammation-01.pdf")`, but I don't see anything that would create or generate the others.  `_episodes_rmd/.gitignore` explicitly ignores `inflammation-01.pdf`, so I'm guessing we can delete all the PDF files.  If you agree, please go ahead and PR something that deletes 'em all.

hdashnow:
In `/data` there are pdf files `inflammation-01.pdf` through to `inflammation-12.pdf`. I suspect these were accidentally committed, and should be removed (also should investigate if the code that creates them there is still in, as that's a bad place for output files!

`ggplot.pdf` looks like lecture slides. I'm not sure of the origin? So best leave them in, but could be stashed in a instructors-extras type folder if that makes things neater?

gvwilson:
Incorporated into #223 - thanks!

naupaka:
The `ggplot.pdf` slides were originally stashed in the intermediate section of the old lesson structure (couple years ago?). I think when the lessons were migrated to the new format we didn't know what to do with the ggplot pdf and just stuck in in here. Is there a place for that sort or material? There should be an `.Rnw` TeX file to generate those beamer slides kicking around somewhere as well. I and @atredennick have been working from that original lesson to teach a [ggplot workshop](https://github.com/atredennick/esa_data_viz_2015) at ESA over the past several years.

gvwilson:
OK, if ggplot.pdf is needed, we'll need to re-introduce it - it can go 
in `files`.

PR #228, opened 2016-08-12
05-cmdline: do not link to a page about kill

keszybz:
There is no need to refer to search through ps aux output, and the
page that was linked wasn't even particularly useful, since 'kill %'
works nicely under most shells. Linked page also suggested using
ctrl-alt-esc, which (I'm guessing) invokes something equivalent to
xkill in some graphical environments. Using this would likely
terminate the terminal or Rstudio, which is definitely not what we
want.


chendaniely:
thanks @keszybz for the changes!

Next time you just need to submit changes for the `.Rmd` file.  We render and build the changes on our end.

PR #231, opened 2016-08-24
Add destination anchor code for each definition list in reference

cmd-ntrf:
During the episodes, we use links to reference definition of certain terms. However, the HTML code produced from the markdown does not associate an id to the definition of the terms, therefore the links point only toward the reference.

This sad reality is true for all of Software Carpentry lessons, but it can be circumvented by adding an anchor tag with the proper id next to the term being defined. This is what I have done in this commit. I would have preferred to find a way to insert a label in Markdown instead of using HTML, but I could not find anything. 

This more a proof of concept, and I would be willing to apply the same modifications to other reference pages if we agreed on the best way to make the reference term link works.

### DIFF
--- a/_episodes_rmd/01-starting-with-data.Rmd
+++ b/_episodes_rmd/01-starting-with-data.Rmd
@@ -36,7 +36,7 @@ knitr_fig_path("01-starting-with-data-")
 
 We are studying inflammation in patients who have been given a new treatment for arthritis,
 and need to analyze the first dozen data sets.
-The data sets are stored in [comma-separated values]({{ site.github.url }}/reference/#comma-separated-values-(csv)) (CSV) format. Each row holds the observations for just one patient. Each column holds the inflammation measured in a day, so we have a set of values in successive days.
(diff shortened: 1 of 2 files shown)


chendaniely:
This seems like a great idea.  and it works!

Do you know if we move the ID tag to before the text it will jump to the line in the reference so that the key is also shown?

for example:

```
argument <a id="argument"/>
```

to

```
<a id="argument"/>argument 
```

cmd-ntrf:
I have rewritten my commit to have the anchor on the line before the text. The key is now shown when clicking on a link.

gvwilson:
The next release of Kramdown (the Markdown parser used by Jekyll on 
GitHub) will automatically insert the ID we want in the `dt` so that we 
can link to it.

cmd-ntrf:
This way simpler than my solution, thanks for the info @gvwilson. The kramdown version supporting this is already out, by I am not sure if Github has updated yet, should I do another pull-request that adds the `{:auto_ids}` to the glossary ? The tag is probably going to be ignored if Github has not updated kramdown to 1.12.0.

gvwilson:
Yes please.

PR #232, opened 2016-08-24
Remove link to sequence in string definition in the glossary

cmd-ntrf:
Sequence is not part of the Glossary for R, only Python.

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -179,7 +179,7 @@ standard output (stdout)
 :   A process's default output stream. In interactive command-line applications, data sent to standard output is displayed on the screen; in a [pipe](#pipe), it is passed to the [standard input](#standard-input) of the next process.
 
 string
-:   Short for "character string", a [sequence](#sequence) of zero or more characters.
+:   Short for "character string", a sequence of zero or more characters.
 
 while loop
 :   A loop that keeps executing as long as some condition is true. See also: [for loop](#for-loop).


chendaniely:
@hdashnow should we add sequence to the R reference?

the Python one is as follows:
```
sequence
A collection of information that is presented in a specific order. For example, in Python, a string is a sequence of characters, while a list is a sequence of any variable.
```

hdashnow:
@chendaniely Yes, I think that's a good idea. The Python definition fits nicely, except maybe as an example we could use a vector instead of list.

cmd-ntrf:
If the term is added to the glossary, it might requires some clarifications. I think in R, this more emphasis on the fact that a sequence has a specific order. Also `sequence` is a function in R, and `seq` is by definition returning a sequence. I am not sure if there is any other function in R that by definition returns a sequence else than `seq`.

hdashnow:
What would you recommend, @cmd-ntrf? Delete or write a new definition?

cmd-ntrf:
I would recommend deleting the link.

The term _sequence_ is used three times in the lesson. Each time it is used to explain the usage of the colon `:` with numbers. The term is sufficiently explained by the context and adding a definition could cause some confusion.

hdashnow:
Sounds good. @chendaniely any objections?

chendaniely:
Sounds great.  I'm convinced!

PR #239, opened 2016-09-26
Activate automatic generation of IDs for terms in reference

cmd-ntrf:
This a follow-up on pull-request #231.

kramdown 1.12.0 now supports the automatic generation of IDs for terms of definition lists. This pull-request activates the feature. http://kramdown.gettalong.org/news.html#kramdown-1120-released

By activating this feature, terms with parentheses have ids that include the text between the parentheses. The pull-request manually fixes the ids for the terms with parentheses as every links in the lesson currently use only the term without the parenthesis content, except for one.

There was only one link that included parentheses, a link to comma-separated value in episode 1. This pull-request removes the `(csv)` from the link.


hdashnow:
Thanks @cmd-ntrf :)

PR #240, opened 2016-09-27
Revert "Activate automatic generation of IDs for terms in reference"

cmd-ntrf:
Reverts swcarpentry/r-novice-inflammation#239

Contrary to what I originally thought, the labels added for some of the terms are not ignored by kramdown 1.11.0 and appear in the webpage. Not good...

We should revert this pull-request until Github updates kramdown to 1.12.0. Sorry about that.


hdashnow:
@cmd-ntrf, no worries. 

What about #242?

cmd-ntrf:
Pull-request #242 will be fine, it is the same as commit 1 in #239. The problem arises when defining specific ids.

PR #243, opened 2016-09-30
Add missing tild before code of increment_me in Reference

cmd-ntrf:


chendaniely:
thanks for the PR @cmd-ntrf

PR #244, opened 2016-10-03
Added SUMMARY function

mchiapello:
I added the "summary" function after the explanation of "min", "max", "mean" and "median" functions. I believe it is a good way to show how to summarise all these functions in just one single command.

Instructor Training: Checkout Procedure for Software Carpentery


chendaniely:
sorry for the delay. but thanks @mchiapello for the PR!

you had some weird `.swp` file that I ended up deleting.
Next time, if you can, you only need to submit the PR for the actual lesson change

PR #246, opened 2016-10-29
better whitespace in lesson: add space between comment character and comment

daattali:


chendaniely:
thanks for the PR @daattali. You only need to submit the changes to the `Rmd` file as we build the site to generate the `md` files manually.

PR #248, opened 2016-11-23
Correcting the section on lists and data frames

dmi3kno:
Intent is to make distinction between matrix and data frame more visible, as well as highlight similarity between list and data frame. Hopefully this will also clarify referencing (subsetting) and will make further concepts easier to grasp (such as nested data frames, *apply family of functions and type simplification/dimension dropping).


dmi3kno:
This pull request can be turned down and closed now. See pull request #252 properly down in own branch

PR #251, opened 2016-11-24
added sapply(df, class)

kdirek:
This is submitted as part of instructor training.

### DIFF
--- a/_episodes_rmd/13-supp-data-structures.Rmd
+++ b/_episodes_rmd/13-supp-data-structures.Rmd
@@ -381,6 +381,7 @@ dat
 > * `ncol()` - number of columns
 > * `str()` - structure of each column
 > * `names()` - shows the `names` attribute for a data frame, which gives the column names.
+> * `sapply(dataframe, class)` - shows the class of each column in the data frame
 {: .callout}
 
 See that it is actually a special list:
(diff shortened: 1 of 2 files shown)


chendaniely:
Hi @kdirek thanks for the PR!

while I agree that this is a useful dataframe function (I do this all the time),
it might be more complicated of a command when compared to the other commands listed.
I feel like this can be a comment or tip when we talk about `apply` or something similar.

@hdashnow, thoughts?

chendaniely:
Also just realized that dataframe columns are stored as a `list`, so we'd also need to distinguish between `lapply` and `sapply`

kdirek:
@chendaniely, I had the same thought on where to place this. For me the decision to include it here and not within the [loops/apply lesson](https://github.com/swcarpentry/r-novice-inflammation/blob/2017.02/_episodes_rmd/15-supp-loops-in-depth.Rmd) was because I felt it more closely aligns to the learning objectives for this lesson and we have already introduced `apply` in the  [first lesson](https://github.com/swcarpentry/r-novice-inflammation/blob/2017.02/_episodes_rmd/01-starting-with-data.Rmd). The information from `sapply(df,` `class)` can also be obtained using `str()` but I would argue the output is more accessible and as a result, more likely to feed into their working practice as it does for you and I.

As for `lapply`and `sapply`, that distinction is briefly outlined in the [loops/apply lesson](https://github.com/swcarpentry/r-novice-inflammation/blob/2017.02/_episodes_rmd/15-supp-loops-in-depth.Rmd) but we could definitely flesh it out a bit with examples.

chendaniely:
@kdirek I've thought about where to place this.
I  I agree with you here putting it in this lesson.

I realized I actually try to show this as soon as possible when I'm teaching because it's such a useful snippet to code.

PR #253, opened 2016-11-26
Fixing inflammation.*csv pattern in 04-cond

jangari:
Pattern is a regex, not a glob.

### DIFF
--- a/_episodes/04-cond.md
+++ b/_episodes/04-cond.md
@@ -564,7 +564,7 @@ Now we can save all of the results with just one line of code:
 
 
 ~~~
-analyze_all("inflammation*.csv")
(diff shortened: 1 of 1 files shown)


jangari:
Ping @hdashnow @chendaniely

bbolker:
This is the same as https://github.com/swcarpentry/r-novice-inflammation/pull/262 but better explained. (It might also be a good idea to add one line somewhere explaining the regex *briefly*, or at least saying that it's magic and pointing users to resources elsewhere ... this regex/`pattern` argument appears first, in passing, on line 282 ...)

chendaniely:
Thanks for the correction. I fixed the source documentation in #268

PR #254, opened 2016-12-01
Add information about variable naming conventions.

oneilsh:
Doing this one as part of my Software Carpentry Checkout procedure.

Newcomers to R are often confused by the variety of naming conventions they encounter, especially if they come from a language where . is semantically meaningful. A quick digression into best practices for naming can both instill good habits and clear up some of the confusion.


chendaniely:
Good point. thanks for submitting the PR!
as people are learning more languages, clarifying the `.` is really important.

PR #255, opened 2016-12-15
New svg diagrams for variables, refactoring subsetting presentation

plantarum:
# Overview

I'm submitting this pull-request as part of my instructor training, which began earlier this fall.

## Three changes here:
- new svg diagrams for the discussion of variables
- re-organized presentation of indexing data frames. As part of this,
removed all references to 'slices', which are a Python thing that is
similar to, but not identical, to indexing and subsetting in R.
- commented out the callout on forcing conversion, which doesn't make sense
to me and suggests that R is worse than it really is in this area.

Also a few typos and other minor corrections.

# Slices
I understand why you'd use that term to explain subsetting to someone coming from Python. But I think there are a few good reasons to avoid it:

- that term isn't used in the standard R documentation, so it won't show up when students search for it in the manuals. There is a function base::slice.index, but it's not doing the same thing. There is also a function `slice` in the Bioconductor package IRanges, but that won't be present until users install it, if they need it.
- slices in Python are similar enough to indexes in R to be misleading.
    - R indexes are inclusive, Python are inclusive only at the bottom. i.e., x[1:3] in R is x[1:4] in Python
    - The `:` operator in R requires a value on both sides: x[1:] is valid in Python, not R
    - negative values in Python slices count from the end; in R they exclude elements

So I don't think we gain much from forcing the term `slice` into service here, and it may cause confusion for those users who know what it means in Python.

# Forcing conversion
I left this section in, but commented it out. I think the explanation is somewhat inaccurate, and will leave the user with the impression that a standard operation in R (selecting a data frame row) will have a different outcome on different computers, with no real reason. I don't think the example data set will cause this problem on any recent R version. The underlying issue is how data frames are constructed: each column is a vector, and there may be columns of different types in the same data frame. That could, in some cases, lead to problems with type coercion. 

I think if we present that issue, it ought to be dealt with head-on, not with an aside that provides a magical fix without context. Instead of offering the advice "if this doesn't work, try this invocation instead", we should explain why it doesn't work, so the user can reason about the issue themselves.

Anyways, I hope I'm not being too strident with my first contribution. I welcome any discussion, criticisms, corrections or even outright rejection!

Best,

Tyler


plantarum:
Regarding coercion of rows to numeric, I posted the following to the mailling list:

> I stand corrected! 
>
> I have had issues with inconsistent (among functions) type coercion
before. Some of these issues have been resolved over time, and I assumed
this was another case of that. However, with some trivial testing, I
find that's not the case. I found the following situation on R 3.3.2:
>
>- `min()` and `max()` call primitive (i.e., C) code, and work as
  expected on data frames (and data frame rows, which are actually
  data frames)
>- `rowMeans()` explicitly converts data frames with `as. matrix()`, and
  so works as expected
>- `sd()` explicitly converts data frames to `numeric()`, and works
  as expected
>- `mean()` does *not* do any coercion, and fails with a warning on data
  frames (and rows)
>
>Which means the message in the lesson is basically sound: sometimes R
functions will treat data frame rows as vectors, and sometimes they
don't, and there's no a priori way to know which is which or why!
>
>With that in mind, I'll think about ways to improve the original callout
to clarify this, if I can.

[Original thread on the mailing list](http://lists.software-carpentry.org/pipermail/discuss/2016-December/004945.html)

chendaniely:
@hdashnow 
how should we then deal with the `slice` in the `references` section of the lesson?

chendaniely:
thanks @plantarum the changes look fantastic.
I agree with your discussion about `slices`. we just now need to discuss what we should do with the `references` page

plantarum:
I can rewrite the 'slices' definition, using the word 'subset' instead. Hopefully in the next week or so?

chendaniely:
Thanks for the update! this is great!

I've fixed all the merge conflicts (sorry it took me so long) and I've opened a separate issue in #290 about the force conversion discussion.

PR #270, opened 2017-03-28
various small fixes

bbolker:
some stylistic (sorry, feel free to ignore): also typos, missing links, tiny clarifications ...


README.md
```diff
@@ -11,13 +11,13 @@ Maintainers:
...
 perform a data analysis. R is used to teach these skills because it is a
 commonly used programming language in many scientific disciplines. However, the
-emphasis is not on teaching every aspect of R, but instead the focus is on
+emphasis is not on teaching every aspect of R, but instead on
 language agnostic principles like automation with loops and encapsulation with
 functions (see [Best Practices for Scientific Computing][best-practices] to
-learn more). In fact, this lesson is a translation of the [Python version][py],
-and the lesson is also available in [MATLAB][].
+learn more). This lesson is a translation of the [Python version][py],
+and is also available in [MATLAB][MATLAB].
```
bbolker: missing link

_episodes_rmd/01-starting-with-data.Rmd
```diff
@@ -430,7 +430,7 @@ We'll learn why this is so in the next lesson.
 > > ## Solution
 > > ~~~
 > > whichPatients <- seq(2,40,2)
-> > whichDays <- c(1:5)
+> > whichDays <- 1:5
```
bbolker: c() is redundant (can't think of an argument for including it?)

_episodes_rmd/03-loops-R.Rmd
```diff
@@ -120,7 +120,8 @@ for (variable in collection) {
 
 We can name the [loop variable]({{ page.root }}/reference/#loop-variable) anything we like (with a few [restrictions][], e.g. the name of the variable cannot start with a digit).
 `in` is part of the `for` syntax.
-Note that the body of the loop is enclosed in curly braces `{ }`.
+Note that the condition (`variable in collection`) is enclosed in parentheses,
```
bbolker: clarification about parens required around loop condition

_episodes_rmd/04-cond.Rmd
```diff
@@ -152,7 +152,7 @@ sign(0)
 sign(2/3)
 ```
 
-Note that when combining `else` and `if` in an `else if` statement (similar to `elif` in Python), the `if` portion still requires a direct input condition.  This is never the case for the `else` statement alone, which is only executed if all other conditions go unsatisfied.
+Note that when combining `else` and `if` in an `else if` statement, the `if` portion still requires a direct input condition.  This is never the case for the `else` statement alone, which is only executed if all other conditions go unsatisfied.
```
bbolker: doesn't seem helpful to make analogy to Python (most students will be unfamiliar)

_episodes_rmd/01-starting-with-data.Rmd
chendaniely: Yes it is redundant, but keeping the `c()` serves as another visual cue to denote a vector.
My concern is that new learners may get confused when we simplify the code too much

bbolker: Hmmm. I don't agree (I think this just muddies people's mental models), but OK.  If you wanted to keep it even simpler, you could use `seq(1,5)` here - but then students wouldn't see the colon operator, which is useful (although dangerous) and ubiquitous.

bbolker:
sorry about conflicts.  They're pretty trivial, though (I tried to fix a paragraph in the best_practices document two different times ...)

_episodes_rmd/01-starting-with-data.Rmd
chendaniely: I'll move this to another PR/Issue.
I can see how it goes both ways. I'm almost tempted to make a small comment that `c(1:5)` and `1:5` are the same thing.

I'll fix the conflicts and merge everything else.

PR #271, opened 2017-03-28
tweaks to best practices

bbolker:
this pull request disappeared (I think) when I reorganized branches on my fork to submit changes from a branch, rather than from the gh-pages branch on my fork [as requested; sorry I didn't do this first time around].

Various comments here, many stylistic (which you should feel free to discard).


_episodes_rmd/06-best-practices-R.Rmd
```diff
@@ -25,33 +25,35 @@ keypoints:
 - "Use version control."
 ---
 
-```{r, include = FALSE}
+```{r source, include = FALSE}
```
bbolker: named all chunks in Rmd (I consider this a "best practice" ...)

_episodes_rmd/06-best-practices-R.Rmd
```diff
@@ -61,44 +63,44 @@ It is best practice to have the user running the script begin in a consistent di
...
+rm("interim_object") # Removes only the object itself and not necessarily the memory allotted to it
 gc() # Force R to release memory it is no longer using
 ls()  # Lists all the objects in your current workspace
 rm(list = ls()) # If you want to delete all the objects in the workspace and start with a clean slate
 ```
 
-11. Don't save a session history (the default option in R, when it asks if you want an `RData` file). Instead, start in a clean environment so that older objects don't remain in your environment any longer than they need to. If that happens, it can lead to unexpected results.
+11. Don't save your workspace (the default option in R, when it asks if you want to "Save workspace image [y/n/c]?"). Instead, start in a clean workspace without old objects cluttering it. Leftover objects from previous sessions can lead to unexpected, hard-to-debug results. Do *not* put `rm(list=ls())` (which removes all objects in your current workspace, as shown in the previous code example) at the top of your code, as this is a trap for other users who might `source()` or copy-and-paste your code in the course of their R session. Instead, restart R when you want to start fresh.
```
bbolker: Important to distinguish between *session history* and *workspace*.  (I may accidentally have repeated some of these edits in a separate pull request ...)  Comment about `rm(list=ls())` is an opinion of mine, but I think (??) fairly widely shared.  Maybe worth an issue/discussion.

_episodes_rmd/06-best-practices-R.Rmd
```diff
@@ -61,44 +63,44 @@ It is best practice to have the user running the script begin in a consistent di
...
 
-12. Wherever possible, keep track of `sessionInfo()` somewhere in your project folder. Session information is invaluable because it captures all of the packages used in the current project. If a newer version of a package changes the way a function behaves, you can always go back and reinstall the version that worked (Note: At least on CRAN, all older versions of packages are permanently archived).
+12. Wherever possible, keep track of `sessionInfo()` somewhere in your project folder. Session information is invaluable because it captures all of the packages used in the current project. If a newer version of a package changes the way a function behaves, you can always go back and reinstall the version that worked (Note: At least on CRAN, all older versions of packages are permanently archived). For more complex projects, you may want to use the [packrat](https://CRAN.R-project.org/package=packrat) package.
```
bbolker: mention packrat

_episodes_rmd/06-best-practices-R.Rmd
naupaka: I am also in agreement with this perspective. +1 to make this change.

naupaka:
I've reviewed these changes and am +1 to merge.

PR #283, opened 2017-05-26
Modifies lesson adding changing column names section.

Raphaelaggio:
https://github.com/swcarpentry/r-novice-inflammation/compare/gh-pages...Raphaelaggio:AddingColumnHeaders?expand=1

### DIFF
--- a/_episodes_rmd/01-starting-with-data.Rmd
+++ b/_episodes_rmd/01-starting-with-data.Rmd
@@ -240,6 +240,27 @@ dim(dat)
 
 This tells us that our data frame, `dat`, has `r nrow(dat)` rows and `r ncol(dat)` columns.
 
+At the moment, the columns of our data frame are named as they were in the csv file `inflammation-01.csv`. We can find the names of `dat` using the command `names`:
+
+```{r}
+names(dat)
+```
+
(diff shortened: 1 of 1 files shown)


chendaniely:
hi @Raphaelaggio Thanks so much for the PR.

While I agree and think we should add the part about `names(dat)` in our data.
I think for a beginner lesson, the use of `gsub` is too much and adds another point for our instructors to cover during an already packed lesson.

My recommendation is to keep only the first part that shows how to get column names using `names`.

_episodes_rmd/01-starting-with-data.Rmd
```diff
@@ -240,6 +240,27 @@ dim(dat)
 
 This tells us that our data frame, `dat`, has `r nrow(dat)` rows and `r ncol(dat)` columns.
 
+At the moment, the columns of our data frame are named as they were in the csv file `inflammation-01.csv`. We can find the names of `dat` using the command `names`:
+
+```{r}
+names(dat)
+```
+
+We know that each row represents one patient and each column represents one time point. It is a good programing practice to use column names that indicate as good as possible the nature of the data. In our case, columns represent time points. Thus, let's modify our column names to T1, T2, T3 and so on. For this, we will use a comman called `gsub`:
```
katrinleinweber: As @chendaniely recommended (I agree), please remove this section and below about changing the column names.

However, your suggestion for a better column naming could be implemented in the raw data files, and highlighted here.

katrinleinweber:
Dear @Raphaelaggio,

because this PR has been open for so long, it has become out of sync a lot. If you are still interested in the actual changes, please consider opening another PR with a branch based off the current `master`.

Thank you, and kind regards!

PR #286, opened 2017-06-23
Subsetting more data solution

AlisaPostma:
Provided a solution with an explanation to the "Subsetting more data" problem in 01-starting-with-data.


chendaniely:
thanks for the solution an PR! :+1:

PR #287, opened 2017-06-27
Update 01-starting-with-data.md

iainsgillis:
## Amend solution for Slicing and Re-Assignment

The exercise presents data for 60 patients (i.e., rows) over 40 days (i.e., columns). Modify `to` argument of `seq` function to number of rows, and transpose the indices in the subset `[` function.

The current solution operates on a subset of the even-numbered days for patients 1 through 5, rather than the even-numbered patients for days 1 through 5.

From Extract.data.frame {base} in the R Documentation, s.v. 'Details':
> When `[` and `[[` are used with two indices (`x[i, j]` and `x[[i, j]]`) they act like indexing a matrix: `[[` can only be used to select one element. Note that for each selected column, `xj` say, typically (if it is not matrix-like), the resulting column will be `xj[i]`, and hence rely on the corresponding `[` method, see the examples section.

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -755,10 +755,11 @@ We'll learn why this is so in the next lesson.
 >
 > > ## Solution
 > > ~~~
-> > whichPatients <- seq(2,40,2)
-> > whichDays <- c(1:5)
+> > whichPatients <- seq(2,60,2)  # i.e., which rows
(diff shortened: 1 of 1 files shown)


chendaniely:
thanks @iainsgillis for catching this! :+1:

chendaniely:
I just caught this mistake when I was generating the lesson.

Next time make the changes to the `rmd` files, as the maintainers will be generating the `md` files for the website.

Thanks for the work!

PR #298, opened 2017-10-16
adding a str()

AnaBVA:
I suggest adding a str() function in order to see how the columns are factors. It can be a 'tip' for checking data when you don't know what is happening in R.


AnaBVA:
sure, thanks

diyadas:
I just noticed that the changes were made to the .md, not the .Rmd file - this is likely to cause issues down the line. I'm not sure how to proceed - @katrinleinweber what are your thoughts?

katrinleinweber:
@AnaBVA & @diyadas: Please have a look and merge :-)

PR #307, opened 2017-12-05
missing word "code"

stragu:
I assume "code" is expected here (which would include "script").


katrinleinweber:
Yes, reads more logical now, thank you :-)

PR #297, opened 2017-09-30
Change variable names in temperature functions for consistency

colinnavin:
Please delete the text below before submitting your contribution. 

---

Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.  

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  

---

PR #320, opened 2018-01-31
follow-up to PR #297: clarify temperatures

katrinleinweber:
I failed to notice the `=` for assignments in #297, so this:

1. reverts only those, 
1. percolates @colinnavin's smart variable renames into the other source files, and 
1. renames the `fahr_…` functions as discussed.


ErinBecker:
Thanks @katrinleinweber  for putting this together. This looks good to me.

PR #322, opened 2018-02-01
correct indices (fixes #308)

### LINKED ISSUE #308: Subsetting More Data (challenge in episode 1) - explanation incorrect
In challenge _Subsetting More Data_ episode http://swcarpentry.github.io/r-novice-inflammation/01-starting-with-data/, the answer is correct but the explanation contradicts what is taught earlier in episode:

Quote:
> Answer: 3 `max(dat[5, 3:7])`
> Explanation: You want to extract the part of the dataframe representing data for patient 5 from days three to seven. In this dataframe, patient data is organised in columns and the days are represented by the rows. Subscripting in R follows the [i,j] principle, where i=columns and j=rows. Thus, answer 3 is correct since the patient is represented by the value for i (5) and the days are represented by the values in j, which is a slice spanning day 3 to 7.

I think this should be:
Explanation: You want to extract the part of the dataframe representing data for patient 5 from days three to seven. In this dataframe, patient data is organised in rows and the days are represented by the columns. Subscripting in R follows the [i,j] principle, where i=rows and j=columns. Thus, answer 3 is correct since the patient is represented by the value for i (5) and the days are represented by the values in j, which is a slice spanning day 3 to 7.

katrinleinweber:
Thanks to @ccronje for 🦅-👁 'edly spotting and reporting this issues :-) This is the correction you meant, right?

### DIFF
--- a/_episodes_rmd/01-starting-with-data.Rmd
+++ b/_episodes_rmd/01-starting-with-data.Rmd
@@ -465,7 +465,7 @@ We'll learn why this is so in the next lesson.
 > >  
 > > Answer: 3
 > >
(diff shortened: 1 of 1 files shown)


ccronje:
Yes, thanks for fixing @katrinleinweber!

Issue #313, opened 2018-01-03
incorrect working directory in supplementary lesson (reading and writing csv files)


diyadas:
The opening to [this section](http://swcarpentry.github.io/r-novice-inflammation/11-supp-read-write-csv/) instructs us to change our working directory to `~/swc` and from there read `data/car-speeds.csv`. If we're following the [setup instructions](http://swcarpentry.github.io/r-novice-inflammation/setup/), the `data` folder containing `car-speeds.csv` is actually located in `~/Desktop/r-novice-inflammation/`.

katrinleinweber:
Hello @diyadas! Yes, [that seems to be a left-over](https://github.com/swcarpentry/r-novice-inflammation/blame/gh-pages/_episodes_rmd/11-supp-read-write-csv.Rmd#L36), when [the initial `setwd` was updated](https://github.com/swcarpentry/r-novice-inflammation/blame/gh-pages/_episodes_rmd/01-starting-with-data.Rmd#L61). 🦅 👀 😄 

Do you want to submit a PR updating `11-supp…` with the correct path?

katrinleinweber:
Cool! In the cases where you are addressing an issue with a commit, here's a neat trick to let GitHub automatically close the issue: https://help.github.com/articles/closing-issues-using-keywords/ ;-)

PR #323, opened 2018-02-01
fixed wrong wd in 11-supp

diyadas:
Updated the working directory in 11-supp-read-write-csv.md to be consistent with the rest of the lesson. Fix for #313 .


katrinleinweber:
Thank you, and have a nice weekend :-)

Issue #329, opened 2018-02-12
issue with SVG files


fmichonneau:
Several SVG files are not rendering correctly, especially in Firefox. 

![image](https://user-images.githubusercontent.com/5502922/36121323-cabbb25a-1013-11e8-841f-f8093bbe4b5c.png)

it might be worth regenerating them to ensure they display correctly across browsers

plantarum:
I just noticed this. I created those images, and at the time they worked as expected for Firefox. But I can confirm that's no longer the case, although they do still look fine on Chrome. I'll take a look at the original files.

plantarum:
I think I've fixed the issue in #330

fmichonneau:
thanks for taking care of this @plantarum. I took a quick look at the files in #330, and they do seem to render well in firefox.

PR #330, opened 2018-02-13
Fixed svg files that didn't render properly in Firefox

plantarum:
I remade a few svg files that weren't rendering properly in firefox. See #329


katrinleinweber:
Thanks for taking the time reporting, and fixing this, @fmichonneau and @plantarum :-)

PR #332, opened 2018-02-28
Edit to command line lesson

AgentKumara:
Edit suggestion to check.R. The original example code tried to calculate the dimension of the filename to obtain the number of rows and columns. This has been edited to calculating the dimension on the file instead.

Suggested edit to the lesson text under the section 'Handling Standard Input' to clarify that `sep = “\n”` was only present in the second `cat`, so that the text prints on the same line.


katrinleinweber:
Hello @AgentKumara and thanks for this update suggestion :-)

PR #333, opened 2018-03-05
changed period to underscore in recommendations for style #294

### LINKED ISSUE #294: suggestion: ".mat" to "_mat"
I suggest that in lesson https://github.com/swcarpentry/r-novice-inflammation, episode 06-best-practices-R, the following:
"For example, name all matrices something ending in .mat"
be changed to:
"For example, name all matrices something ending in _mat"

This is only a tentative suggestion. The motivation is to avoid encouraging the practice of using dots in object names, except in classes. I am somewhat conflicted, because dots are nice and easy to read. But I thought ".mat" might be confusing on two fronts -- 1. if learners undertake lessons in python at the same time, the use of the dot to call methods might get confusing, and 2. if learners are also taking a matlab class, the ".mat" file extension might be a source of confusion.

alexwhan:
Very minor PR to fix #294

### DIFF
--- a/_episodes_rmd/06-best-practices-R.Rmd
+++ b/_episodes_rmd/06-best-practices-R.Rmd
@@ -100,7 +100,7 @@ source("my_genius_fxns.R")
 ```
 
 ### Other ideas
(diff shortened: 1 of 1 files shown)


katrinleinweber:
Thanks for implementing that suggestion :-)

PR #337, opened 2018-03-09
Introduce regex with digits instead of wildcard

katrinleinweber:
[shell-novice/04-pipefilter](https://github.com/swcarpentry/shell-novice/blob/gh-pages/_episodes/04-pipefilter.md#using-wildcards) introduces `.*` as a wildcard for file-names. On the surface, this seems to be the same as in `list.files(… pattern = "…")`, but that needs to be a regular-expression, in which the `.` matches many more characters than the `.` before the file extension in the shell.

Hence, and in order to prepare future learning of REs, I suggest to restrict the RE here to the number sequence that the example files actually have.

- [x] an explanatory sentence each for `[0-9]` and `{2}` is needed.

### DIFF
--- a/_episodes_rmd/03-loops-R.Rmd
+++ b/_episodes_rmd/03-loops-R.Rmd
@@ -318,7 +318,13 @@ Let's test out running our `analyze` function by using it on the first three fil
 
 
 ```{r loop-analyze, fig=FALSE}
-filenames <- list.files(path = "data", pattern = "inflammation-[0-9]{2}.csv", full.names = TRUE)
+filenames <- list.files(path = "data", 
+                        pattern = "inflammation-[0-9]{2}.csv", 
+                        # This pattern is a regular expression that matches:
+                        # a) the static part of the filenames,
+                        # b) the variable part (two of the digits 0 to 9), and
+                        # c) the static file type CSV.
+                        full.names = TRUE)
 filenames <- filenames[1:3]
 for (f in filenames) {
   print(f)
--- a/_episodes_rmd/03-loops-R.Rmd
+++ b/_episodes_rmd/03-loops-R.Rmd
@@ -318,7 +318,7 @@ Let's test out running our `analyze` function by using it on the first three fil
 
 
 ```{r loop-analyze, fig=FALSE}
-filenames <- list.files(path = "data", pattern = "inflammation.*csv", full.names = TRUE)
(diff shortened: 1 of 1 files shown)


ErinBecker:
@katrinleinweber - sorry for my delayed review on this. Since I'm not familiar with the r-novice-inflammation curriculum in general, I don't feel that I can give good advice about whether an introduction to REs belongs here or not. Maybe @naupaka, @jcoliver, @mawds, @tomwright01, @aurielfournier, @mondorescue or @fmichonneau can offer assistance as Maintainers of other R lessons in the Carpentry stack.

naupaka:
I agree with the conceptual problem here. But: introducing regular expressions is a whole thing and it might be a bit of a tangent at this point in the lesson (just a guess - I'm not sure on that point since I am not as familiar with the inflammation R lessons). In this particular case, we could sidestep the whole issue by just changing to `list.files(path = "data", pattern = "inflammation", full.names = TRUE)`, since there are no non-csv files with "inflammation" in their name.

katrinleinweber:
I agree with your "tangent" warning, but "sidestepping" around our own examples files wouldn't make a good impression IMHO, would it? Because a) filename patterns such as these are common and thus b) learners are likely to encounter them in their real-world tasks (dates, times, iterators from some measurement device, etc.), I find the intro of `[0-9]` and `{N}` sufficiently useful, logical within this example's context, while still being minimal enough.

diyadas:
Hey all, I could go either way on this. Seeing as the concept of regular expressions is taught in the unix shell lesson, and in recent workshops at least, the unix lesson is taught alongside R, it might not be that confusing to incorporate. 

I would definitely prefer a change. As it is, I found the current pattern confusing when I first looked at it, because I kept mentally swapping `*` and `.`

P.S. [Episode 7 of shell](https://swcarpentry.github.io/shell-novice/07-find/) links to a [SWC tutorial on regex](http://v4.software-carpentry.org/regexp/index.html) that might be worth linking to from this lesson as well, so the details of regex don't overtake the lesson but leaving an "external" resource for those who want to learn more.

katrinleinweber:
I have reservations about linking to the (possibly out-dated, see referenced commit above) `v4` site. If the reply there is "still up-to-date", we'll use that.

Otherwise, how about indenting the comment to the appropriate reg-ex part, see my above commit?

katrinleinweber:
In [that case](https://github.com/swcarpentry/shell-novice/issues/758), I prefer the indented mini-legend ;-)

PR #338, opened 2018-03-09
Make folder-nature explicit in path arguments?

katrinleinweber:
As the learners saw in [shell-novice/02-filedir](https://github.com/swcarpentry/shell-novice/blob/gh-pages/_episodes/02-filedir.md), a trailing `/` differentiates as directory/folder from a file.

This is something we could reiterate here, but apparently at the cost of the console output duplicating the slash: `data//inflammation-01.csv`. Worth it?

### DIFF
--- a/_episodes_rmd/03-loops-R.Rmd
+++ b/_episodes_rmd/03-loops-R.Rmd
@@ -284,8 +284,8 @@ Since no pattern is specified to filter the files, all files are returned.
 So to list all the csv files, we could run either of the following:
 
 ```{r}
-list.files(path = "data", pattern = "csv")
-list.files(path = "data", pattern = "inflammation")
+list.files(path = "data/", pattern = "csv")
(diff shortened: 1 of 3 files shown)


jdblischak:
Thanks for getting this discussion started, @katrinleinweber.

Have you had learners that were confused by the first argument to `list.files`? I think specifying the name of the argument `path =` makes it explicit that it is a path and not a filename. Furthermore, the default argument is `path = "."` and not `path = "./"`.

I do not have a strong opinion on this either way. But if we add the trailing slash to indicate the argument is a directory, then I think we should save typing by not naming the argument, e.g.

```
list.files("data/", pattern = "csv", full.names = TRUE)
```

katrinleinweber:
Good point! Your suggestion would reiterate what they learned about named, unnamed and positional arguments :-)

PR #341, opened 2018-03-13
roxygen2 is on CRAN. No need to install from github

haozhu233:
roxygen2 is on CRAN. No need to install from github
Also added a sentence about `@export`.


katrinleinweber:
Hello, thanks for this update and welcome to this Carpentry lesson :-)

Issue #236, opened 2016-09-21
command line lesson (lesson 5) will differ depending on who is compiling the document


chendaniely:
If I `make lesson-rmd`  I will always get a change for the some of the chunks because it calls functions like `sessionInfo()`.  The difference is from my computer being a different OS than the current lesson (which seems was generated from OSX)

Is there a better way to handle this?

This might just be an issue us maintainers need to deal with...

katrinleinweber:
This is bugging me as well now, also because the messages are in German. Makes no sense to commit _those_. I'm ignoring it manually for now.

However, in order to avoid this as an error source, _and_ keep the OS version numbers in the examples up-to-date, one of us should update it before a lesson release?

diyadas:
@fmichonneau : @katrinleinweber and I are resolving a lot of issues today :) What do you think of this one?

katrinleinweber:
Maybe related to #376.

fmichonneau:
I can set up the repo to be built and deployed from Travis so it will always be the same environment that generates the markdown files.

diyadas:
I like that solution. Thanks @fmichonneau !

fmichonneau:
I should be able to set it up by the end of the week. Do you think it will be possible to get the PR merged/closed by then? Enabling Travis requires switching the default branch, and it would affect open PRs.

diyadas:
Oh! @katrinleinweber should confirm - I know she was planning some work during these next two weeks which includes updating and merging one of the open PRs.

katrinleinweber:
Yes, I have nothing planned during September, as I'll be on vacation. As we discussed in the last call, we completed #298 & I verified that we have no headings deeper than a single level per episode. So, go Travis :-)

diyadas:
Hi @fmichonneau - any chance we could make the switch this week? I'd love to go ahead before more PRs roll in.

Thanks!

katrinleinweber:
I'm on holiday until the end of September. Please feel free to go ahead.

fmichonneau:
working on this now. I'll let you when it's completed and functional.

fmichonneau:
The transition to Travis is now complete. A few important considerations:

- **The default branch for this repository is now `master` (this will affect existing pull requests)**.
- After each push to `master` Travis takes the content of the repository, knits the episodes, and push the content of the repository to the `gh-pages` branch. From there GitHub takes care of rendering the website at the usual location.
[...]

PR #360, opened 2018-05-24
Teach about roxygen2 earlier (closes #309)

### LINKED ISSUE #309: Teach roxygen2 skeleton instead of free-form comments
Hello!

Since the function comments in [lesson 2](http://swcarpentry.github.io/r-novice-inflammation/02-func-R/) & [3](http://swcarpentry.github.io/r-novice-inflammation/03-loops-R/) cover two thirds of the [roxygen2 skeleton](https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html) `@param`, `@return` & `@examples`, I wonder whether it would be worthwhile to shift the teaching to that standard and relegate the unstructured docu examples to the interactive / script-like lesson content?

What do you think?

Cheers :-)

PS: A little side effect on the lesson would be the need to upgrade "we often use RStudio, but it is not required" to an _actual_ prerequisite, because of the `Shift+Alt+Cmd+R` shortcut / code tool.

katrinleinweber:
This provides a suggestion to implement #309.

- [ ] trim town `08` to only repeat a bit of the earlier roxygen2 info

### DIFF
--- a/_episodes_rmd/01-starting-with-data.Rmd
+++ b/_episodes_rmd/01-starting-with-data.Rmd
@@ -135,7 +135,7 @@ We can treat our variable like a regular number, and do arithmetic with it:
 
 > ## Commenting
 >
-> We can add comments to our code using the `#` character. It is useful to
+> We can add [comments]({{ page.root }}/reference/#comment) to our code using the `#` character. It is useful to
 > document our code in this way so that others (and us the next time we
 > read it) have an easier time following what the code is doing.
 {: .callout}
--- a/_episodes_rmd/02-func-R.Rmd
+++ b/_episodes_rmd/02-func-R.Rmd
@@ -284,7 +284,7 @@ We have one more task first, though: we should write some [documentation]({{ pag
 
 A common way to put documentation in software is to add "informal" comments like this:
 
-```{r}
+```{r center-commented, eval=FALSE}
 center <- function(data, desired) {
   # return a new vector containing the original data centered around the
   # desired value.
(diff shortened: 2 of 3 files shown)


katrinleinweber:
Thanks for the review @lvclark!

I'm sorry that I missed this for 3 weeks.

> I made the R-package-structure.svg figure, so I can revise it for this version if we want to put it back in.

I'd be happy to include it in if you send a PR against the `309...` branch :-) Let's wait for two weeks, because that's when I will have taught this episode.

> […] show the learners roxygenise the first time around, rather than giving them a mystery […]

I'll see how this challenge works out in the workshop. I'm open to switching it around, just not shortly before it.

lvclark:
@katrinleinweber sounds good, I'm on vacation right now anyway, so I can remake the diagram once I get back and once you see if you need to revise the lesson any more.

katrinleinweber:
Hi @lwjohnst86! Since you taught that part in the end, do you remember the learner reaction to the challenge of `roxygenise()`-ing the [documentation written earlier](https://github.com/TIBHannover/FAIR-R/blob/gh-pages/_episodes_rmd/03-func-R.Rmd#technical-details-of-rs-help-page-format).

lwjohnst86:
@katrinleinweber I don't recall there being any real reaction to it. But then again, I don't recall bring up the reason why `?` didn't work. I think I just mentioned "Ctrl-Shift-D" for `devtools::document()` fairly quickly after showing the roxygen documentation to talk how to generate the documentation. Hope that helps!

katrinleinweber:
Yes, if it was your intuition to show this quickly, it's probably better to adjust the lesson that way. Thanks :-)

katrinleinweber:
This is getting a bit out of hand. I'll try resetting `08-packaging` and updating it only minimally.

katrinleinweber:
Not useful, because `analyse()` is challenged to be packaged, but the main part of the episode is about [the temperature conversions, which are not being touched by my changes to `02-func-R`](https://github.com/swcarpentry/r-novice-inflammation/pull/360/files#diff-63af2862f6a0eb4a35d6ff63ffd7c122R37).

I feel that the de-sync has grown too much. Shall we close this, @lvclark & @diyadas? I'm grateful for your advice along the way, and we have the variant [TIBHannover.GitHub.io/FAIR-R](https://tibhannover.github.io/FAIR-R/) in which that advice is integrated from the ground up to focus on function documentation, packaging and testing.

diyadas:
I think closing makes sense. I'd want to defer work until we hear back from the CAC at least on #391 and I'd like to make the switch to Travis soon, so we'd likely find it simpler to start over at that point (#236).

lvclark:
Sorry for the delay in my responding.  I'm a little too overwhelmed with other stuff at work right now, and not sufficiently familiar with other changes to this lesson, to have an opinion at the moment.  But if you would still like me to remake my diagram please let me know.

katrinleinweber:
OK. Let's leave the branch, though.

PR #362, opened 2018-06-12
edit contributing guidelines for clarity

diyadas:
1. edit Rmds not mds
2. examples of "significant" changes

(I was not at my best at that hour and it took four commits -- please squash and merge if you agree with the proposed changes.)


diyadas:
Added a new commit FYI, not sure if GitHub automatically notifies of new commits to existing PRs.

katrinleinweber:
Yes it does :-) Thanks!

PR #368, opened 2018-06-16
Include links.md in all episodes

maxim-belkin:
Add `{% include links.md %}` to all episode files

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -978,3 +978,5 @@ The maximum value rises and falls perfectly smoothly, while the minimum seems to
 >
 > Create a plot showing the standard deviation of the inflammation data for each day across all patients.
 {: .challenge}
+
+{% include links.md %}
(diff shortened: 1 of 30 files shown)


katrinleinweber:
Hi Maxim! Thanks for all these PRs (#367, #369 & #370) :-)

I was thinking about the best way to place links for a while as well, but in my tests `{% include links.md %}` at the end of `.md` files didn't work. I assume you ran `make serve` and tested this change?

Only adding the same line to the `.Rmd` files seems to work for me.

maxim-belkin:
Hi Katrin,

Honestly, I don't care where these links are included. The ultimate goal is to have TravisCI working and for that we need to fix all errors produced by `make lesson-check`. Currently, `lesson-check` checks that `links.md` is included on the last line of `.md` files (see `check_reference_inclusion` in `lesson_check.py`)

Let's ask @rgaiacs if this is true for `r-novice-inflammation` lesson as well or if we should update the `lesson_check.py`

katrinleinweber:
I'm pretty sure they have to be there, but since we are generating the `.md` out of `.Rmd`, I am also sure they are needed there as well.

PR #371, opened 2018-06-17
06-best-practices-R.md: Fix link to SWC lessons' page

maxim-belkin:
Fix URL. Currently, `links.md` does not define `[lessons]`, so we have to use actual URL.

### DIFF
--- a/_includes/links.md
+++ b/_includes/links.md
@@ -18,6 +18,7 @@
 [jekyll-windows]: http://jekyll-windows.juthilo.com/
 [jekyll]: https://jekyllrb.com/
 [jupyter]: https://jupyter.org/
+[lessons]: https://software-carpentry.org/lessons/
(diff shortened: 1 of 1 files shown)


katrinleinweber:
Thanks for the note about this :-) I think it will be more elegant to introduce your suggestion in a #368-compatible way. It seems that the easiest way to do this, is if you revert 0a52cf0.

maxim-belkin:
I think changes to `links.md` have to be done in carpentries/style repo.

I'm sure it is alright to make changes locally first, but let's submit a PR upstream as well and see what Raniere and François say :)

rgaiacs:
> I think changes to links.md have to be done in carpentries/style repo.

Yes, please.

katrinleinweber:
OK, moved to https://github.com/carpentries/styles/pull/303

PR #376, opened 2018-06-20
WIP: Make site to align factor order with Rmd

katrinleinweber:
For a while now I noticed the following diff after each `make site/serve`, but I don't understand what it results from. I don't see what recent change in [12-supp-factors.Rmd](https://github.com/swcarpentry/r-novice-inflammation/blame/7400b93450253fadbb0ee1263bb950fb4ca4f06a/_episodes_rmd/12-supp-factors.Rmd#L146) could cause this, and neither in the [`.md` version](https://github.com/swcarpentry/r-novice-inflammation/blame/2c56ed752a52160213109d11735aa35ca2886f95/_episodes/12-supp-factors.md#L254-L264) before 428c315.

Could this be some OS-specific behavior, or due to an R upgrade (to v3.4.0, because I saw this even before upgrading to 3.5.0 just a few days ago).

@diyadas: When you `make site` on gh-pages, do you also see changes in the same files an in this PR?


diyadas:
I wonder since the issue is due to levels of a factor, if there's something about the locale as well. IIRC, yours may not be `en_US`.

katrinleinweber:
OK, thanks! I had at least good intuition to always discard those changes :-D

I have `[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8` as well, though. Will dive into `.Profile` & remove the attachments.

```
R version 3.5.0 (2018-04-23)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS High Sierra 10.13.5

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] testthat_2.0.0 magrittr_1.5   usethis_1.3.0 

loaded via a namespace (and not attached):
[1] compiler_3.5.0  backports_1.1.2 R6_2.2.2        tools_3.5.0     yaml_2.1.19    
[6] knitr_1.20      rlang_0.2.1    
```

fmichonneau:
I think there is, unfortunately, a part of randomness in PNG file generation, and even with the same system and exactly identical images, the file content may differ...

katrinleinweber:
In case of the actually drawn pixels, yes, but here the factors are reordered, so the bar/box order and number also changes.

katrinleinweber:
Closing to prepare switch to Travis, see #236.

PR #381, opened 2018-07-07
typo fix

steltenpower:


katrinleinweber:
Thank you :-) In this case, there was a little mistake though: you edited the `.md` file which we normally auto-generate from the `.Rmd` source. I copied your correction to that and merged both together, see https://github.com/swcarpentry/r-novice-inflammation/commit/522bf55d303a57312cad330cd62a355a02bd3770#diff-d1cab001de9a60cdcf48b89e8049301d.

The above mention of `50915a0` is therefore out-dated.

PR #386, opened 2018-07-24
this addresses fix #379. Content has been added to the matrix section…

### LINKED ISSUE #379: Suggestion: use typeof() to demonstrate that matrices are extensions of vectors and dataframes are extensions of lists
(This is part of my instructor checkout)

In the Data Types and Structures sections, the `class` and `typeof` functions are introduced. Under the matrix section, matrices are described as extensions of numeric and character vectors. Data-frames as described as special lists. It might be helpful to include `typeof()` and `class()` on a matrix and on a data-frame to illustrate that while the class might say matrix/data-frame, under the hood, they are indeed vectors and lists. It also clears up the difference between the two functions as in the notes, the one example where both `typeof` and `class` are used together on the same object (a character vector), the same result is returned (character).

aabarug:
… to show the difference between the `class()` and `typeof()` functions and to show the matrices are vectors. The data-frame section was not modified as it uses `is.list()` in combination with `class()` to show this as well.

### DIFF
--- a/_episodes_rmd/13-supp-data-structures.Rmd
+++ b/_episodes_rmd/13-supp-data-structures.Rmd
@@ -267,6 +267,16 @@ m
 dim(m)
 ```
 
+You can check that matrices are vectors with a class attribute of `matrix` by using `class()` and `typeof()`.
+
+```{r}
+m <- matrix(c(1:3))
+class(m)
+typeof(m)
+```
+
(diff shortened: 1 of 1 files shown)


aabarug:
Awesome! :)

diyadas:
Thanks to both of you! I've been traveling but it looks great and I've merged.

PR #396, opened 2018-10-10
Align margin order in fig/r-operations-across-margins.svg with text

katrinleinweber:
Hi! I'm always confused about the `apply(.., MARGIN = ...)` and I want to try whether swapping the order in the figure helps. Please compare visually: [current](https://github.com/swcarpentry/r-novice-inflammation/blob/master/fig/r-operations-across-margins.svg) vs. [swapped](https://github.com/swcarpentry/r-novice-inflammation/blob/537424e5bb317e894f7689389864a4125c12e10c/fig/r-operations-across-margins.svg).

### DIFF
--- a/fig/r-operations-across-margins.svg
+++ b/fig/r-operations-across-margins.svg
@@ -9,15 +9,15 @@
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    version="1.2"
-   width="513.76117"
-   height="269.27908"
-   viewBox="0 0 14499.482 7599.6541"
+   width="331.29922"
+   height="268.1983"
+   viewBox="0 0 9350.0004 7569.1521"
    preserveAspectRatio="xMidYMid"
    clip-path="url(#presentation_clip_path)"
    xml:space="preserve"
    id="svg12375"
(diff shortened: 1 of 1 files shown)


katrinleinweber:
Pinging @jdblischak because he seems to have created the diagram ;-)

jdblischak:
> he seems to have created the diagram ;-)

@katrinleinweber Thanks! But I can't take credit for that diagram. It was derived from an early version of [python-operations-across-axes.svg](https://github.com/swcarpentry/python-novice-inflammation/blob/v5.3/fig/python-operations-across-axes.svg) (r-novice-inflammation is a translation of python-novice-inflammation).

>  swapping the order in the figure helps

What exactly did you swap? I've stared at the before and after for a while now, but must admit I don't see it. In both versions, the arrows for the rows are on the left and the arrows for the columns are on the right. I apologize if I am missing an obvious change.

jdblischak:
> Oh shit. I think Travis already deployed the swapped version to https://swcarpentry.github.io/r-novice-inflammation/fig/r-operations-across-margins.svg. That's why you saw the same, swapped version in different places. Fixed in OP now.

Well now I feel better about myself now for not being able to tell them apart :-)

Anyways, I like your changes to the figure. I agree with you that for R it is natural to think about the row operations first and then the columns to correspond to the values for `MARGIN`.

PR #406, opened 2018-12-13
equation render not working

evezeyl:
- the equation render does not work - I guess because of what (typo?) I removed (but new to that - because learning)- At least you know that this does not work as expected at this line. Best regards


katrinleinweber:
Since the only affected file was a .gif, I suspected it had something to do with how GitHub pages renders images. I replaced it with a .png version.

Thank you again for this bug report :-)

katrinleinweber:
Sorry about this! 8ce3640 didn't solve the problem after all. What do you think about this solution?

katrinleinweber:
Because that solution reduces complexity on our end, and because we can always fall back to the PNG version in the future, I'm merging this now.

@evezeyl: Thanks again for alerting us about this! All the best for 2019 🎉 😃

Issue #399, opened 2018-10-15
Permission denied error when rendering Rmarkdown file


statkclee:
I'm trying to translate R Novice Rmd file into Korean.
At the final stage for the HTML production, I've got the following error,

`kclee@MacBook:~/swc/r-novice-kr$ sudo make lesson-md`

`make: bin/knit_lessons.sh: Permission denied`
`make: *** [_episodes/01-starting-with-data.md] Error 1`

Is there any walkaround?

katrinleinweber:
Hello! This is probably related to our recent clean-up (https://github.com/swcarpentry/r-novice-inflammation/commit/98825367d8ecf555e567cd8538f403f259940313). Does the process not create the `_episodes` folder by itself? Does `mkdir _episodes` and repeating the process prevent this error?

katrinleinweber:
It shouldn't be. [`make lesson-md`](https://github.com/swcarpentry/r-novice-inflammation/blob/98825367d8ecf555e567cd8538f403f259940313/Makefile#L88-L92) should just work. I'm on it!

katrinleinweber:
OK, I bisected through the history but could not reproduce this problem, at first. It does not seem to be an issue with the code.

Thus, it might be with your copy of `knit_lessons.sh`. Please try the following in a terminal/shell:

1. `ls -lg bin/knit_lessons.sh` -- If the returned permission set is something like `-rw-r--r--` (no `x`!) the file is not marked as e`x`excutable.
2. You can make it so with `chmod +x bin/knit_lessons.sh`.

Both commands assume that you `cd`'ed into the `r-novice-inflammation` root folder. Please try this and let me know whether it runs through then.

katrinleinweber:
PS @diyadas: The [Travis failure](https://travis-ci.org/swcarpentry/r-novice-inflammation/builds/441060276#L1706) reported for this repo currently, seems to be different, although the error message contains `01-starting-with-data.md` as well. The site still works, as does the download link from the `files` location (see #398).

I can't investigate this week, so I suggest the not-so-great option of waiting for another commit or PR to trigger Travis again.

diyadas:
Thanks for the heads up. I suspect that it does need the` _episodes` folder to be a part of the repo? 

@fmichonneau can you confirm, off-hand?

katrinleinweber:
OK, that can be easily tested.

katrinleinweber:
As usual, you were rigth :-)

[...]

PR #408, opened 2019-01-02
Improve lesson structure for maintainers

katrinleinweber:
This summarizes a few minor things I noticed during #399 and elsewhere. Please see my inline-comments below.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -83,6 +83,7 @@ defaults:
 # Files and directories that are not to be copied.
 exclude:
   - Makefile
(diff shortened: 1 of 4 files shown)


_config.yml
```diff
@@ -83,6 +83,7 @@ defaults:
 # Files and directories that are not to be copied.
 exclude:
   - Makefile
+  - r-novice-inflammation.Rproj
```
katrinleinweber: I found it annoying to always have the 2nd copy of that file in `_site`.

PR #409, opened 2019-03-01
Update 06-best-practices-R.Rmd

jmablans:
Suggested change in file reference in sample code for constancy. 

In the beginning of the script an `input_file` and an `output_file` are defined.
Only the `input_file` is used further in the script.

Please delete the text below before submitting your contribution.


diyadas:
Welcome @jmablans and thanks for this contribution!

(@katrinleinweber and others: just for clarification, data was being written with an undefined variable `results_file` instead of the `output_file` defined earlier in the lesson.)

katrinleinweber:
Nice catch! Thank you @jmablans :-)

PR #412, opened 2019-03-27
Swap elementwise operators to R-recommendation

### LINKED ISSUE #410: Teaching '&' instead of '&&'.
In episode 4 we teach element-wise AND and OR comparisons.

> We can also combine tests. An ampersand, &, symbolizes “and”. A vertical bar, |, symbolizes “or”. & is only true if both parts are true:

However, for control flow `&&` and `||` are preferred:

> & and && indicate logical AND and | and || indicate logical OR. The shorter form performs elementwise comparisons in much the same way as arithmetic operators. The longer form evaluates left to right examining only the first element of each vector. Evaluation proceeds only until the result is determined. **The longer form is appropriate for programming control-flow and typically preferred in if clauses.** [Emphasis mine; from the `Logic {base}` help() page. ] 

Is this an oversight in the lesson, or is it intentional?

katrinleinweber:
Hello @reedacartwright & my apologies for the late reply!

Thank you for raising this issue :-) Looking at that episode's [`git blame` info](https://github.com/swcarpentry/r-novice-inflammation/blame/master/_episodes_rmd/04-cond.Rmd), I don't think there was any decission about this. Please have a look at the changes I suggest here.

Things I consider **important before merging**:

- [x] read entire episode to find other text that might need to be updated
- [x] review other episodes &(&) update as well ;-)

Closes #410.

### DIFF
--- a/_episodes_rmd/04-cond.Rmd
+++ b/_episodes_rmd/04-cond.Rmd
@@ -9,14 +9,14 @@ questions:
 objectives:
 - "Save plot(s) in a PDF file."
 - "Write conditional statements with `if` and `else`."
-- "Correctly evaluate expressions containing `&` ('and') and `|` ('or')."
+- "Correctly evaluate expressions containing `&&` ('and') and `||` ('or')."
 keypoints:
 - "Save a plot in a pdf file using `pdf(\"name.pdf\")` and stop writing to the pdf file with `dev.off()`."
 - "Use `if (condition)` to start a conditional statement, `else if (condition)` to provide additional tests, and `else` to provide a default."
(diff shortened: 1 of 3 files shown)


reedacartwright:
@katrinleinweber I grepped the other lessons and I could not find any other ones that used the elementwise operators. I think this is good to go.

katrinleinweber:
Thanks for checking! I'm thinking about switching all `test`s for `condition` or `conditionals`. What do you think @diyadas?

diyadas:
Sorry, just realized you might not have gotten a notification, but I'm in agreement.

katrinleinweber:
Thank you :-)

PR #414, opened 2019-04-09
Clarify question about rescale()'s default arguments

### LINKED ISSUE #413: Unresolved question about default arguments?
Please see: https://github.com/swcarpentry/r-novice-inflammation/blob/21a0ee0c7eaf71ccd4136b7691ade9aefaf739fe/_episodes_rmd/02-func-R.Rmd#L503-L535

Do you all feel that the "… functions always behave the same way?" question is resolved, by the `min` & `max` results below. Should there be some explanatory text?

katrinleinweber:
Closes #413. See discussion there.

I'm still a bit worried about this challenge not having any explanatory text in the "Solution". If anyone knows examples of incorrect implementations, plus the test inputs that uncovered them, those might be a good addition here. Please comment or commit those!


katrinleinweber:
Cool, thanks :-)

Issue #415, opened 2019-04-16
June 2019 Lesson Release checklist


fmichonneau:
If your Maintainer team has decided not to participate in the [June 2019 lesson release](https://carpentries.topicbox.com/groups/maintainers/T61c75ac1a8a89538/timeline-and-process-for-june-lesson-release), please close this issue.

To have this lesson included in the 18 June 2019 release, please confirm that the following items are true:

- [x] Example code chunks run as expected
- [x] Challenges / exercises run as expected => @katrinleinweber went through all non-`supp`
lementary material
- [x] Challenge / exercise solutions are correct => #424  & #427
- [x] Call out boxes (exercises, discussions, tips, etc.) render correctly => @katrinleinweber checked for left-over `>`, `{:`, etc. in HTML & **found none**
- [x] A schedule appears on the lesson homepage (e.g. not “00:00”)
- [x] Each episode includes learning objectives
- [x] Each episode includes questions
- [x] Each episode includes key points
- [x] Setup instructions are up-to-date, correct, clear, and complete => 86a5148 
- [x] File structure is clean (e.g. delete deprecated files, insure filenames are consistent) => done in #395, #398, #408 & others
- [x] Some Instructor notes are provided => [_extras/guide.md](https://github.com/swcarpentry/r-novice-inflammation/blob/master/_extras/guide.md)
- [x] Lesson links work as expected => 001ca45 repairs 1 broken link

When all checkboxes above are completed, this lesson will be added to the 18 June lesson release. Please leave a comment on carpentries/lesson-infrastructure#26 or contact Erin Becker with questions (ebecker@carpentries.org).

diyadas:
We've discussed and we'd like to be included in the June lesson release. Will work through the checklist.

katrinleinweber:
I'll also run `make lesson-check-all` and remove all warnings.

katrinleinweber:
Regarding "File structure is clean": Can we remove the whole boilerplate folder?

fmichonneau:
it's not needed by the template. Depending on the number of lessons that will be involved in the release, I'm planning to migrate some/all to using remote themes. With this approach, none of the template files would be part of the repository anymore and updates to the template will happen with every push.

katrinleinweber:
Sounds good! #418 will remove those files for now and in case any later merge conflicts appear because of that, I take responsibility to for resolving them.

diyadas:
[...]

PR #417, opened 2019-04-18
Improve variable & function names

katrinleinweber:
Just a bunch of renamings that hopefully make the purpose of each renamed element clearer.

- [x] also update https://github.com/swcarpentry/r-novice-inflammation/blob/master/_episodes_rmd/14-supp-call-stack.Rmd#L130
- [x] search other episodes

### DIFF
--- a/_episodes_rmd/02-func-R.Rmd
+++ b/_episodes_rmd/02-func-R.Rmd
@@ -209,11 +209,12 @@ Real-life functions will usually be larger than the ones shown here--typically h
 ### Testing and Documenting
 
 Once we start putting things in functions so that we can re-use them, we need to start testing that those functions are working correctly.
-To see how to do this, let's write a function to center a dataset around a particular value:
+To see how to do this, let's write a function to center a dataset around a
+particular midpoint:
 
 ```{r}
-center <- function(data, desired) {
-  new_data <- (data - mean(data)) + desired
+center <- function(data, midpoint) {
+  new_data <- (data - mean(data)) + midpoint
   return(new_data)
 }
 ```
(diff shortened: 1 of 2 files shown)


_episodes_rmd/02-func-R.Rmd
```diff
@@ -285,11 +285,11 @@ We have one more task first, though: we should write some [documentation]({{ pag
 A common way to put documentation in software is to add [comments]({{ page.root }}/reference.html#comment) like this:
 
 ```{r}
-center <- function(data, desired) {
+center <- function(data, around) {
   # return a new vector containing the original data centered around the
-  # desired value.
+  # around value.
```
diyadas: I agree that this change makes sense and I like this PR a lot. The only point of confusion is that this now reads "original data centered around the around..." which might take a moment.

Perhaps: original data centered around the value of "around" ? A few more words.

katrinleinweber: Yeah 😅 Or `mid_point` as the parameter name?

diyadas: I like that a lot!

katrinleinweber: Done :-) Thanks for confirming!

katrinleinweber:
I found another occurence of `fence` thanks to https://github.com/swcarpentry/r-novice-inflammation/issues/424#issuecomment-489684648 and renamed `outside` to `edges` because that seems more meaningful for the 1st & last element of a vector ;-)

katrinleinweber:
I would like to have this included in #415, so I'll merge now. Feel free to revert if you disagree.

Issue #393, opened 2018-09-10
Adding options(stringAsFactors = FALSE) [episode 11]


duytpm16:
Hi,

Would it be worth mentioning to learners that having the line of code, options(stringAsFactors = FALSE), at the beginning of a script is another solution to prevent R from converting strings to factors?

Particularly in 11. Reading and Writing CSV files or 12. Factors (under Default Behavior)

I find this trick very useful because I don't have to manually set stringAsFactors to FALSE when I'm reading in multiple files that I know do not contain categorical variables. 

Duy Pham

diyadas:
Hi Duy, thanks for your comment!

Episode 11 already has an entire section on stringsAsFactors = FALSE in the reading of csv files, so I would put it there. I don't want to keep adding material due to length, but it is a useful trick and I think one sentence at the end of the section would be fine. The trick is to avoid adding an entire section on options, but maybe another sentence linking to where to find more information about options would solve this.

I'd be willing to review a pull request on this topic, if you or anyone else wants to work on it.

katrinleinweber:
Hi @duytpm16: Are you interested in starting the PR Diya suggested or should someone else do it?

taddallas:
Issue is a bit stale and perhaps is ready to be closed? It will also become a non-issue in R 4.0+, as the shift towards stringsAsFactors=FALSE default will be incorporated by then (hopefully). https://developer.r-project.org/Blog/public/2020/02/16/stringsasfactors/index.html

HaoZeke:
Yup, this is part of the discussion on migration, and shouldn't be an issue at the moment. Closing for now. Thanks @taddallas @diyadas and @katrinleinweber !

Issue #276, opened 2017-04-22
Latest RStudio Update Uses readr As Default Point-And-Click Importer


Geeketics:
In the latest RStudio update the readr package has been linked to the "Import Dataset" button in RStudio, this used to be the utils package. This can cause problems downstream, particularly with subsetting.

I know SWC lessons are all about making people learn to code things, instead of point-and-click, but it seems important to add a note to all the R lessons that warns the learner that the point-and-click "Import Dataset" button in RStudio may not do what they expect it to and that this could cause problems later on. Basically it would be emphasising the need to actually type in the code, rather than trying to take a point-and-click short cut.

chendaniely:
thanks for the heads up. do you know where we should add this comment?

I feel like it needs to be mentioned multiple times in the lesson:

1. while importing data
2. when we talk about `as.data.frame`

I don't know off the top of my head what the differences between the `tibble` object and the regular `data.frame` object is...

bbolker:
The main/most import difference I'm aware of is that tibbles do *not* drop dimensions by default as other R rectangular data structures do: in particular, `tbl[,"foo"]` is a tibble rather than a vector (one should use `tbl[["foo"]]` to get a vector). That being said, it's probably sufficient to tell beginning students "tibbles usually but **not always** act the same as data frames: if necessary, use `as.data.frame()` ..."

katrinleinweber:
Wouldn't this instance even be a starting point to start converting the lesson to tibbles? Continuing to teach a non-default (compared to the most popular R-IDE) might cause more problems than it's worth.

pauljohn32:
I'm not an RStudio user/enthusiast.  I think it imposes the particular opinions and packages of the authors onto the users in a restrictive way.  Given that, you may not be surprised with my counter argument: if the software imposes judgments that make you feel you need to teach about different code content,  you should consider getting a different software tool.  You would not put up with a C++ editor that imposes somebody's library of routines for floating point calculations on you. Would you?  In that case, I'd say use a different editor.  R existed long before RStudio and many of us who write R packages don't use RStudio at all.

bbolker:
[...]

PR #419, opened 2019-04-21
Adding options(stringsAsFactors=FALSE)

katrinleinweber:
Revisiting #393, I noticed that @duytpm16's commit was not in any PR, yet. It:

> add['s] a little section on options(stringsAsFactors = FALSE) under  `The `stringsAsFactors` Argument`[and] add['s] a short line of code as well to show how it works. […]

I am a bit skeptical of `options()` ([similarly to `setwd()` or `rm(list = ls())`](https://twitter.com/JennyBryan/status/997004754759778305)) to be honest, but it at least works only during the current session.

The alternative would be to solve @duytpm16's usecase

> don't … manually set `stringAsFactors` to `FALSE` when … reading in multiple files

by teaching to wrap the `read…` function into a custom one, similarly to the ["Functions to Create Graphs" challenge](https://swcarpentry.github.io/r-novice-inflammation/02-func-R/#functions-to-create-graphs). See esp. [my last review comment](https://github.com/swcarpentry/r-novice-inflammation/pull/419#discussion_r277174511), but I wanted to give this suggestion a fair chance & discussion.

### DIFF
--- a/_episodes_rmd/11-supp-read-write-csv.Rmd
+++ b/_episodes_rmd/11-supp-read-write-csv.Rmd
@@ -115,6 +115,19 @@ carSpeeds$Color
 ```
 
 That's better! And we can see how the data now is read as character instead of factor.
+ 
+ 
+ 
+Alternatively, if we want to prevent characters from converting to factors, we can change R's default behavior by using `options(stringsAsFactors=FALSE)`. This will change the default value of the 'stringsAsFactors' argument to FALSE in all functions that utilize this parameter without us having to do so manually. 
+
+```{r optionsStringsAsFactorFALSE}
+options(stringsAsFactors=FALSE)
+carSpeeds <- read.csv(file = 'data/car-speeds.csv')
+str(carSpeeds)
+```
+
+See ?options for more details about the function and a list of other default settings that can be changed.
+
 
 
 ### The `as.is` Argument


_episodes_rmd/11-supp-read-write-csv.Rmd
```diff
@@ -115,6 +115,19 @@ carSpeeds$Color
 ```
 
 That's better! And we can see how the data now is read as character instead of factor.
+ 
+ 
+
```
katrinleinweber: Do we need this many newlines above & below? I think 1 or 2 are enough separators.

_episodes_rmd/11-supp-read-write-csv.Rmd
```diff
@@ -115,6 +115,19 @@ carSpeeds$Color
 ```
 
 That's better! And we can see how the data now is read as character instead of factor.
+ 
+ 
+ 
+Alternatively, if we want to prevent characters from converting to factors, we can change R's default behavior by using `options(stringsAsFactors=FALSE)`. This will change the default value of the 'stringsAsFactors' argument to FALSE in all functions that utilize this parameter without us having to do so manually. 
+
+```{r optionsStringsAsFactorFALSE}
+options(stringsAsFactors=FALSE)
```
katrinleinweber: Does RStudio's `Code` > `Reformat…` leave the `=` without surrounding spaces? If not, please update L121 as well.

_episodes_rmd/11-supp-read-write-csv.Rmd
```diff
@@ -115,6 +115,19 @@ carSpeeds$Color
...
+carSpeeds <- read.csv(file = 'data/car-speeds.csv')
+str(carSpeeds)
+```
+
+See ?options for more details about the function and a list of other default settings that can be changed.
```
katrinleinweber: Would this be a good place to mention that each `options()` change is valid only during the current R session?

Since this touches on using [`.Rrofile` to change the defaults permanently](https://www.r-bloggers.com/fun-with-rprofile-and-customizing-r-startup/), and on rather teaching the tidyverse way (see #276), I'm worried that this change risks bloating the lesson.

_episodes_rmd/11-supp-read-write-csv.Rmd
diyadas: (Reformat code adds spaces, so please update.)

_episodes_rmd/11-supp-read-write-csv.Rmd
diyadas: Agreed. I might even suggest a change above, to line 121, and abandon this part:

>If we want to prevent characters from converting to factors, we can change R's behavior in the current session by using `options(stringsAsFactors = FALSE)`. See `?options` for more details about the function. Note that any changes made using `options` will not be reflected in a new R session, unless the command is re-run.

katrinleinweber:
Closing due to inactivity & because `options` would be a [new concept](https://github.com/swcarpentry/r-novice-inflammation/blob/master/CONTRIBUTING.md#what-not-to-contribute). Will reopen upon continuation of the above discussion.

PR #421, opened 2019-04-23
Upgrade 06-best-practices with more references or even practices?

katrinleinweber:
Besides, this PR drafts & discusses a few related upgrades, namely suggestsions to include more references to enable learners to read more about the reasoning behind each item we suggest. I think we should avoid including _more_ suggestions, unless they are short and backed by references.

My goal here is to:

- [x] suggest references for the existing best practices

to exemplify the above.

### DIFF
--- a/_episodes_rmd/06-best-practices-R.Rmd
+++ b/_episodes_rmd/06-best-practices-R.Rmd
@@ -115,7 +115,7 @@ source("my_genius_fxns.R")
 
 3. Don't repeat yourself--automate! If you are repeating the same code over and over, use a loop or a function to repeat that code for you. Needless repetition doesn't just waste time--it also increases the likelihood you'll make a costly mistake!
 
-4. Keep all of your source files for a project in the same directory, then use relative paths as necessary to access them. For example, use
+4. Keep all of your source files for a [project in the same directory][rstudio-project], then use relative paths as necessary to access them. For example, use
 
 ```{r relpath, eval=FALSE}
 dat <- read.csv(file = "files/dataset-2013-01.csv", header = TRUE)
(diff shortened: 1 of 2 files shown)


Afrah-Khairallah:
https://www.r-bloggers.com/r-code-best-practices/

katrinleinweber:
I think we should either be a bit more specific, linking each of our items to another resource, or agree to replace our list with a particular resource or list.

diyadas:
Greg has an article that could be linked here: https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005510

HaoZeke:
I'm not sure we ought to be linking to external content like that. Perhaps it might be better to create summary pages based on the external sources and link to them at the end of the article? Like "inspired by the guidelines at ...". That way we would be able to deal with content being modified out of the scope of our control.

katrinleinweber:
> create summary pages

That expisode is exactly that already ;-)

> That way we would be able to deal with content being modified out of the scope of our control.

For a video, that's an unlikely worry. For text-pages: OK, but then I'd rather switch to archived version. Like https://web.archive.org/web/20190103191213/https://support.rstudio.com/hc/en-us/articles/200532077 for `[rstudio-git]`.

HaoZeke:
Archive pages are a great idea, @katrinleinweber. They're easier to maintain as well.

katrinleinweber:
Not necessarily _easier_, because you introduce a few additional steps per link. You only get piece-of-mind in not needing to check for external updates. However, I think @diyadas and me did that only when we noticed (or where pointed to) actually 404-ing links.

HaoZeke:
True but by pinning them once with IA, we never need to worry about 404s. This is actually more relevant if we use less popular sites which are not periodically updated on IA in the first place.

katrinleinweber:
> [...] never need to worry about 404s

Our [CI finds them](https://github.com/swcarpentry/r-novice-inflammation/blob/c00ea02897d6617ea1277ab69f0f95900d20cb18/.travis.yml#L10) ;-)

It's no longer my prerogative to decide, but my advice is to avoid inserting _any_ extra step. The risk of problematic external changes is low (check [`links.md`'s blame history](https://github.com/swcarpentry/r-novice-inflammation/blame/master/_includes/links.md)), so it's sufficient to deal with them when they arise / are reported (["YAGNI" principle](https://en.wikipedia.org/wiki/YAGNI)).

HaoZeke (approved):
Moving forward with this now.

PR #429, opened 2019-05-13
Turn hard-coded path into default argument (fix #305)

### LINKED ISSUE #305: Analyzing Multiple Data Sets : Challenge
**Challenge: Using Loops to Analyze Multiple Files**

**Current Version**
analyze_all <- function(pattern) {
  filenames <- list.files(path = "data", pattern = pattern, full.names = TRUE)
  for (f in filenames) {
    analyze(f)
  }
}

**Proposed Version: Two Input Parameters**
analyze_all <- function(**_folder_**, pattern) {
  filenames <- list.files(path = **_folder_** , pattern = pattern, full.names = TRUE)
  for (f in filenames) {
    analyze(f)
  }
}

I think the proposed version is only slightly more complex than the current version, conceptually speaking; but useful practically.
Also, it fits well with the more complex "function writing" examples from earlier modules, where two parameters were involved.

katrinleinweber:
Reviewing #305 now, I think that also the comments and the challenge text would need to be updated. What do you think about these :arrow_upper_right: changes @Jerry-Lai?

### DIFF
--- a/_episodes_rmd/03-loops-R.Rmd
+++ b/_episodes_rmd/03-loops-R.Rmd
@@ -348,15 +348,16 @@ Sure enough, the maxima of these data sets show exactly the same ramp as the fir
 
 > ## Using Loops to Analyze Multiple Files
 >
-> Write a function called `analyze_all` that takes a filename pattern as its sole argument
+> Write a function called `analyze_all` that takes a folder path and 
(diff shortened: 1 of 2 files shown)


diyadas:
Looks great to me!

PR #432, opened 2019-05-24
Rephrase slice as subset

### LINKED ISSUE #430: Rephrase "slice" with "subset"
In #255 many such renamings and rephrasings were already made. However, [some instances of `slic`es are left](https://github.com/swcarpentry/r-novice-inflammation/blame/11c258c33e79dd43f78d0d16b5d95f45d91d97c6/_episodes_rmd/01-starting-with-data.Rmd#L377-L469) and these too, should be replaced with `subset`.

katrinleinweber:
Fixes #430.


katrinleinweber:
Cool, thanks!

PR #437, opened 2019-06-21
Update styles to gh-pages:6d015fe

### LINKED ISSUE #436: Labels of code chunks (bash/R/output) in lesson 5, command line
Hi everyone, 

I got a bit confused by the labels of code/output chunks in 
[lesson 5 http://swcarpentry.github.io/r-novice-inflammation/05-cmdline/index.html ](http://swcarpentry.github.io/r-novice-inflammation/05-cmdline/index.html).

For instance the first one is labelled "R" but is bash code; while the 4th one is labelled "output" while it is R code without any output (although not meant to be run in R directly).
I haven't counted exactly, but maybe about a fifth of the chunks seems mislabelled. 

If the problem is real (I may just be super confused!) I am happy to try and submit a pull request correcting it.

I would label as "Bash" (using `{: .language-bash}` ) anything that is meant to be written and run in the command line, label as "R" (using `{: .language-R}`) anything that is meant to be written and run in R, and label "Output" anything that is returned by R or Bash. I am not 100% sure what to do with chunks that show R-code not meant to be run directly, but instead to be saved and called in bash with Rscript later. The "R" label seems more logical to me, but maybe they were meant to look different from chunks to be run directly in R, hence the use of "Output" for them? (If so, it may be good to create a new label type for those?)

Cheers,

Timothee

katrinleinweber:
Closes #436. Due to b50bada80fbefbecd40d5e9eeb3c0fca6cb107e6 in #418, I resolved the conflicts in some boilerplate files by keeping them deleted from this repo.

### To do

`make serve` locally & review visually:

- [x] `aoi`
- [x] `05-cmdline`


katrinleinweber:
Thanks for reporting this, @timotheenivalis and thanks for providing the fix @fmichonneau 	:-) Solves for the main issues, but I wonder still about the `Output`:

image

https://github.com/swcarpentry/r-novice-inflammation/blob/7aa4e9b278c3620fa65e9674bd6a532ad9f6637f/_episodes_rmd/05-cmdline.Rmd#L68-L70

It arguably _is_ output, just of `cat`. If we want to clarify that as well, maybe removing all the `echo=FALSE` would make the rendered sites more logical?

PR #439, opened 2019-07-05
Update line-count.R

jperkel:
Original version incorrectly counts total line numbers if only one filename provided on command line. Removing the `if (length(args) > 1)` check fixes that.

### DIFF
--- a/_episodes_rmd/line-count.R
+++ b/_episodes_rmd/line-count.R
@@ -8,12 +8,12 @@ main <- function() {
       cat(filename)
       cat(" ")
       cat(num_lines, sep = "\n")
-      if (length(args) > 1) {
-        total_lines = total_lines + num_lines
-      }
+      total_lines <- total_lines + num_lines
     }
-    cat("Total ")
-    cat(total_lines, sep = "\n")
+    if (length(args) > 1) {
+      cat("Total ")
+      cat(total_lines, sep = "\n")
+    }
   } else {
     input <- readLines(file("stdin"))
     num_lines <- length(input)


katrinleinweber:
Dear @jperkel,

thank you for notifying us of this bug and suggesting a fix :-) Because I'm on travels, I may need to wait up to next weekend to reproduce & verify both.

Maybe @diyadas can have a look in the mean-time?

Kind regards,
Katrin

diyadas:
Hello! I have also been traveling but I will have a look tonight. Thanks both!

diyadas:
Hi, I managed to take a quick look this morning. This is indeed a bug - thanks for catching it. 

The challenge question only asks for the total number of lines if more than one file is provided, so I think the precise solution would be to wrap the `cat` statements in question with `if(length(args)>1)`. However, I’d also be fine with removing it entirely if the challenge text were updated to ask for a total number of lines across all files provided to `line-count.R`. 

@katrinleinweber do you have a preference?

_episodes_rmd/line-count.R
```diff
@@ -8,9 +8,7 @@ main <- function() {
       cat(filename)
       cat(" ")
       cat(num_lines, sep = "\n")
-      if (length(args) > 1) {
-        total_lines = total_lines + num_lines
-      }
+      total_lines <- total_lines + num_lines
     }
     cat("Total ")
     cat(total_lines, sep = "\n")
```
katrinleinweber: If we moved the `if` around these two `cat`s, the code would work as intended _and_ not show the `Total` for a single input file.

diyadas: Sounds good to me!

katrinleinweber: d2b237c would be my suggestion. @jperkel: Are you OK with that? Feel free to squash these commits locally & `push --force` them again, so that the git history credits you with finding & fixing this bug ;-)

jperkel: Thanks @katrinleinweber. I'm not sure how to do what you are suggesting -- sorry, I'm a GitHub novice. How do I do this? Thanks!

katrinleinweber: You're welcome :-) OK, let's do this! After you `git clone https://github.com/jperkel/r-novice-inflammation/` your repo locally (if you already did that when starting this PR, better clone it again into another parent folde):

1. `cd r-novice-inflammation && git checkout patch-1` moves you into the repo & onto this branch from which you are requesting us to pull.
1. `git rebase -i e1d2addddec940a05a45758f9029fd48064da53c` starts the squashing process
1. in the editor window that opens, you need to replace the `pick` in front of my 2 commits (7956b6 & d2b237) with `fixup` (or `squash` if you also want to change the commit message), then save & close.
1. `git push --force origin patch-1` to overwrite our 3 commits here with the new, single commit.

You can also use [one of these tutorials](https://duckduckgo.com/?q=git+rebase+interactive+tutorial). I'll be happy to help you at any point if you want. Use `git log | grep -E '<ka.*@\w+\.' | sort | uniq` to extract my email address(es) from the git repo ;-)

jperkel: Thanks @katrinleinweber, very helpful! I did as you instructed -- let me know if that didn't work please.
Jeff

katrinleinweber: It did work :-) Welcome to advanced git! In real-live, `rebase -i` is best only used before any `push`, and if you do have to rebase already pushed commits, best notify other parties involved in a PR and use [`--force-with-lease`](https://git-scm.com/docs/git-push#Documentation/git-push.txt---force-with-leaseltrefnamegt).

PR #440, opened 2019-07-31
Convert directly from Fahrenheit to Kelvin

gringer:
The Fahrenheit to Kelvin formula was previously based around converting the temperature first to Celsius, then to Kelvin. This request changes the formula to avoid the intermediate Celsius step (which could be confusing), and adds a reference for that formula (for those who want to give a taste of looking for answers on the Web).

When I consult the 'Fahrenheit' Wikipedia page, the formula for conversion is on the first page. It added a bit of time to my live coding demonstration (less than a minute), but helps to demonstrate how you can use information from the Web in your code.

I've tested this via *make serve* to make sure it renders okay on my own computer:

Jekyll screenshot


katrinleinweber:
Being a European, `32` & `273.15` are not confusing to me at all ;-) I understand that it may be opposite for US-educated learners but I think we should stick to the numbers that arise from grounding this formula on the [SI](https://en.wikipedia.org/wiki/International_System_of_Units).

gringer:
But SI defines temperature in K, not C. The 32 and 273.15 are associated with Celsius, not Kelvin.

katrinleinweber:
459.67 isn't associated with Kelvin either ;-) The metric variant is clearly in the majority, esp. in the sciences. That's what I think the lesson should support and I'm opposed to changing this on the grounds of "could be confusing".

You're right about me overstretching the SI argument, though. My apologies for failing to notice this.

gringer:
459.67 is the zero-point (or in mathematical terms the offset) for 0K in Fahrenheit; it needs to be in there for a conversion directly from Fahrenheit to Kelvin.

katrinleinweber:
@diyadas: What do you think about this?

@gringer: How about _adding_ it as a code comment directly below the Celsius-numbers? That would give learners & instructors the opportunity to pick the variant they want in that moment. The `aside` note could then become a shorter reference to the Wikipedia page.

diyadas:
I remember 273.15 and 32 as constants as an American scientist, for what it's worth. I have not learned a direct Fahrenheit to Kelvin conversion, but rather Fahrenheit to Celsius and Celsius to Kelvin conversion, as it's written in the lesson.

I would not be opposed to a comment as @katrinleinweber has indicated.

gringer:
273.15 and 32 are constants related to *Celsius* and its relationship to Kelvin and Fahrenheit respectively. Celsius first appears [in the next step](http://swcarpentry.github.io/r-novice-inflammation/02-func-R/index.html#composing-functions).

Introducing / explaining Celsius right at the start of the lesson is not necessary to get the point of function definitions across, and makes the function body more complicated than it needs to be.

_episodes_rmd/02-func-R.Rmd
```diff
@@ -39,7 +39,7 @@ Let's start by defining a function `fahrenheit_to_kelvin` that converts temperat
 
 ```{r}
 fahrenheit_to_kelvin <- function(temp_F) {
-  temp_K <- ((temp_F - 32) * (5 / 9)) + 273.15
+  temp_K <- (temp_F + 459.67) * 5 / 9
```
katrinleinweber: Because [my co-maintainer seems to agree](https://github.com/swcarpentry/r-novice-inflammation/pull/440#issuecomment-517583426), please change your F-based variant into a code comment and re-insert the C-based line above.

That probably achieves our common goal of _not confusing_ learners with at-the-moment irrelevant information as a [minimum viable change](https://about.gitlab.com/handbook/values/#minimum-viable-change-mvc).

_episodes_rmd/02-func-R.Rmd
```diff
@@ -49,6 +49,8 @@ The list of argument names are contained within parentheses.
 Next, the [body]({{ page.root }}/reference.html#function-body) of the function--the statements that are executed when it runs--is contained within curly braces (`{}`).
 The statements in the body are indented by two spaces, which makes the code easier to read but does not affect how the code operates.
 
+Aside: the formula here is from [wikipedia](https://en.wikipedia.org/wiki/Fahrenheit#Definition_and_conversion), and defines the zero point for the Kelvin scale as −459.67°F, with a scaling factor of 5/9.
+
```
katrinleinweber: Please merge this aside into the short comment above, as briefly as possible. We don't need to explain the formula itself, because that's not the point of the lesson.

katrinleinweber:
Besides the review comments summarising the discussio so far, I want to point out that [no other Carpentries lesson currently uses the Fahrenheit-based numbers](https://github.com/search?o=desc&q=org%3Acarpentries+org%3Aswcarpentry+org%3Adatacarpentry+org%3Alibrarycarpentry+%22%2A_to_kelvin+%3C-+function%22&s=indexed&type=Code).

Therefore, if you are opposed to minimising this PR, and instead really want to "rejigg this lesson to do the fahrenheit_to_celsius conversion first" (Sounds like a good idea to me :+1:), please provide some objective justification for "is more common". Also, I wonder if in that case, your suggestion should be implemented as a set of coordinated PRs against all lessons, in order to avoid any confusion arising from different example math in different lessons.

What do you think cc @swcarpentry/r-novice-gapminder-maintainers @swcarpentry/python-novice-gapminder-maintainers @swcarpentry/python-novice-inflammation-maintainers?

I'm fine with either keeping this PR small, or turning it into the forum for this discussion.

maneesha:
I love that this discussion is happening because it not only considers what we teach (like writing functions) but also how it can be made to feel relevant to the work of scientists and researchers.

Two things to note -- the Python inflammation lesson includes a [demo to convert Fahrenheit to Celsius](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/_episodes/06-func.md).  I personally like this better than including Kelvin at all, because it remains domain agnostic.  Any person would be familiar the F and C temperature scales while only people in certain scientific fields would be familiar with Kelvin.

gringer:
I also like the flow of the Python inflammation lesson. These lessons eventually use all three scales (for "Composing Functions"), so I don't think we can ignore Fahrenheit or Kelvin completely. With regards to justifying the order, here are Google search result estimates:

"celsius to fahrenheit" - About 502,000 results
"fahrenheit to celsius" -  About 450,000 results
"celsius to kelvin" - About 99,700 results
"kelvin to celsius" - About 86,000 results
"fahrenheit to kelvin" - About 60,200 results
"kelvin to fahrenheit" - About 59,700 results

So if we were to choose the function order based on popularity, it would be C->F, then C->K, then the composed function would be F->K... but that doesn't work, because F->K would require the C->F function to be inverted. The Python inflammation approach (F->C, then C->K, then F->K composed as F->C->K) seems to be a reasonable compromise.

I think that a commented mathematically-equivalent function would be more confusing than leaving the text unchanged. It encourages the distracting question to be raised ("why are the formulas different?" - which actually happened in my demo run through), diverging from the goal of the lesson. Given the prevalence of the Fahrenheit-to-Celsius-to-Kelvin formula, I can see how inconsistent my proposed change would be unless it were changed *everywhere*, which seems like too much of a change for a "my first carpentries pull request".

diyadas:
Leaving aside the content, I’d just like to say that your first Carpentries PR doesn’t need to be a small thing. We’re obviously in the midst of a discussion and I agree that it’s good to get input from other maintainers / synchronize across lessons, but you being a new contributor doesn’t matter in this context. That being said, welcome!

maxim-belkin:
Before picking a side, I'd like to explain the reasons (or at least how I justify it to myself) why we have two functions in the SWC-Py-inflammation lesson: F-to-C and C-to-K.

If we abstract away and think globally there are two main "splits": US--EU and "normal people"--scientists ("normal people" can be defined as scientists at home with their families). Both splits co-exists and intersect: there is a US--EU split among scientists and normal people, and there is "normal people"--scientists split in both US and EU. 

Let's consider the first split: US uses F, Europe -- C. This means that "normal people" use either Fahrenheit or Celsius. Therefore, it makes sense to have F2C or C2F conversion function... to help people understand "the other temperature scale" better. This is fine and dandy but what about Kelvin?

Whenever I hear someone use Kelvin, my "science radar" is activated. Of course, not all science domains use Kelvin: K are mostly used in low-temperature science and when talking about energies (_k<sub>B</sub>T_), but I haven't heard of any geographic location where Kelvins are used officially. So, it seems to make sense to have F2K or C2K (or K2F and K2C) conversion function to help "normal people" understand what "science people" say. 

Now, which function combinations of (F2C, C2F, F2K, and C2K) should we choose?
And this is where swc-py-inflammation is different from this lesson. We chose to talk US-EU and normal people --- scientist language. This lesson "speaks" US--science, science-EU

let's note that our end goal is to teach _scientists_ function reuse: we'd like to show that one function can use other function(s). 

Because of that, we're actually choosing between two pairs:
F2C + C2K  and C2F + F2K.

_episodes_rmd/02-func-R.Rmd
maxim-belkin: @gringer, the formula is, of course, not from Wikipedia. Wikipedia is simply the place that collects information from other sources. I agree with @katrinleinweber in that there is no need to explain the formula in the text.

katrinleinweber: Thanks for d7ee2437e8c3bcf73588b72bf2947fe1324648ff, but still: Is this an added explanation, really needed?

gringer: Not necessarily, it depends on the audience. The direct conversion from Fahrenheit to Celsius (without any intermediate calculations, as in the current Fahrenheit to Kelvin method) does make function a lot more obvious, so an explanation is less necessary. I think that explanations for formulas are useful to give people a concept that they can hang their understanding on.

Formula both with and without explanation are used in [the previous section](http://swcarpentry.github.io/r-novice-inflammation/01-starting-with-data/index.html). When the conversion from kg to lb is first used, it's just stated in code:
>     # weight in pounds
>     2.2 * weight_kg

Whereas a bit further on (a few steps later) the operations are expanded out:

> In words, we’re asking R to look up the value we tagged weight_kg, multiply it by 2.2, and tag the result with the name weight_lb

maxim-belkin: 1. This paragraph switches the discussion from function(s) to _formula_ -- note that text before and after talks about functions, not formula. This is not ideal.
2. If I were picky, I would say that the formula does not _define_ "zero point" (btw, you would have to explain what "zero point" is) -- it defines the rule for converting temperature from one scale to another. Imagine if someone told you that "`w_lb = 2.2 * w_kg` defines zero point for pounds as 0 kg with a scaling factor of 2.2".
3. Lastly, current position of the paragraph makes it appear as this part is essential to understanding how to write functions in R. But it is not, of course. If you feel that it is important to have some explanation -- do that before the discussion of functions begins. This might be helpful/useful: people would first see mathematical expression first and then how wonderful it is in R.  I would suggest doing so in a separate PR, though.

Another note: I'm a Pythonista and can't help but note that the first sentence in the preceding paragraph ("We define... by assigning ... to the output...")  could be a little tiny bit improved. Words define / assign might be new to people. which output?

Yet another note: in the sentence "In this lesson, we'll learn how to write a function...". I'd say "we'll learn how to quickly execute the same set of operations on different data with the help of functions that we can create ourselves." (or something along those lines).

maxim-belkin:
@gringer, please have a look at https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/fig/README.md for some tips on how to clean up SVG files -- see if it can be useful here.

gringer:
I can. Is it okay to make those SVG optimisation changes within this pull request (as a "while I'm here, I might as well fix X" change)?

katrinleinweber:
Everything that affects the temperature-related figures is fine to include here :-)

Changes to _other_ figures should go into a different PR, though.

_episodes_rmd/08-making-packages-R.Rmd
```diff
@@ -66,27 +66,27 @@ Suggestion: organize in a logical manner so that you know which file holds which
 Let's turn our temperature conversion functions into an R package.
 
 ```{r}
-fahrenheit_to_kelvin <- function(temp_F) {
-    # Converts Fahrenheit to Kelvin
-    temp_K <- ((temp_F - 32) * (5/9)) + 273.15
-    temp_K
+fahr_to_celsius <- function(temp_F) {
```
katrinleinweber: Is this renaming/shortening necessary? We teach to ["name and style code consistently"](https://swcarpentry.github.io/r-novice-inflammation/06-best-practices-R/index.html) and expanded the abbreviations for [these](https://github.com/swcarpentry/r-novice-inflammation/pull/297#issuecomment-361623083) [reasons](https://github.com/swcarpentry/r-novice-inflammation/pull/320).

_episodes_rmd/02-func-R.Rmd
katrinleinweber: > Note that I've changed the explanation in 013b394 to remove the Wikipedia link.

I don't find the link problematic at all, because it can blend into any text. The added "Explanation:" paragraph/line does not.

katrinleinweber:
I tested the flow of the episode just now and am I'm quite happy with the change in function order. Thank you :-) Once the two "shortening" and the "explanation" discussions are resolved, I'm fine with merging this.

_episodes_rmd/08-making-packages-R.Rmd
gringer: Because that's how it was in the image. I was changing the functions to match the image.

gringer:
Yes, I'm fine with squash-merging.

There's another niggle related to conventions in the stack episode. It uses the example of two ``temp`` variables to explain why it's useful to have multiple stack frames, which is [again] appears to be something that was pulled from the Python code (and is [still there](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/_episodes/06-func.md)).

However, the R code uses explicit ``temp_F`` and ``temp_K`` variables in the initial function definitions, which makes the stack frame explanation a little bit less obvious ("we have called them ``temp_F`` and ``temp_C``, but sometimes there are functions that have exactly the same variable name").

I'll update the images to match the function definitions (with expanded names). Are there any ideas about the best way to approach the stack frame explanation?

_episodes_rmd/14-supp-call-stack.Rmd
```diff
@@ -22,56 +22,56 @@ knitr_fig_path("14-supp-call-stack-")
...
 
 This final stack frame is always there;
 it holds the variables we defined outside the functions in our code.
 What it *doesn't* hold is the variables that were in the various stack frames.
-If we try to get the value of `temp` after our functions have finished running, R tells us that there's no such thing:
+If we try to get the value of `temp_F` after our functions have finished running, R tells us that there's no such thing:
 
 ```{r, error = TRUE}
-temp
+temp_F
```
katrinleinweber: Well spotted :-) `temp` seems to have been a year-old left over 😞

katrinleinweber:
Thanks for plowing through all this @gringer 	:-)

Any objections, @diyadas? Without reply, I'll merge early next week. cc @maxim-belkin

maxim-belkin:
SVG: unless `<tspan>`s are used for styling purposes, one can change all 
```svg
<text><tspan></tspan></text>
```
to `<text></text>`. `<tspan>`s allow one have multiple lines of text within a single `<text>` element.

Regardless of that -- great work, @gringer! 🎉 

And thank you for inviting me and all other maintainers, @katrinleinweber.

katrinleinweber:
> `<text><tspan></tspan></text>`

I'd rather have that in a different PR.

diyadas:
Sorry for my inactivity; things got busy and I wasn’t able to follow up. For what it’s worth, I agree with the course of action taken here. Thanks for seeing this through!

Issue #296, opened 2017-09-19
Change of class for data frame subsets by column


gosianow:
In Programming with R, Addressing Data, there is a line of code that shows the class of `dat`:
`class(dat)`
which is a `"data.frame"`. Maybe just for completeness, it would be worth mentioning that when subsetting by column, the class of what we get changes. For example, for `dat$Gender`
`class(dat$Gender)` is `"character"`.

katrinleinweber:
As in #295, please feel free to send a PR. Your suggestion makes sense, so I'd be happy to review a specific "implementation" / addition to the lesson :-)

PR #446, opened 2019-08-30
Addressing issue #296 Change of class for data frame subsets by colum…

ggrimes:
I changed example dim(dat) to str(dat) to seed the idea that data frames are made up of different classes of vector.  In addition, I specifically, address issue #296 by getting students to inspect the class of two different columns.

### DIFF
--- a/_episodes_rmd/10-supp-addressing-data.Rmd
+++ b/_episodes_rmd/10-supp-addressing-data.Rmd
@@ -45,7 +45,7 @@ Lets take a look at this data.
 class(dat)
 ```
 
-R has loaded the contents of the .csv file into a variable called `dat` which is a `data frame`. A data frame is a list of variables of the same number of rows with unique row names, given class "data.frame".
+R has loaded the contents of the .csv file into a variable called `dat` which is a `data frame`.
 
 We can compactly display the internal structure of a data frame using the  structure function `str`.
 
@@ -149,7 +149,7 @@ We usually use the `$` operator to address a column by name
 dat$Gender
 ```
 
-When we address a data frame by a single column, R will return a vector of that column class and not a data frame.
+When we extract a single column column from a data frame using the `$` operator, R will return a vector of that column class and not a data frame.
 
 ```{r }
 class(dat$Gender)
--- a/_episodes_rmd/10-supp-addressing-data.Rmd
+++ b/_episodes_rmd/10-supp-addressing-data.Rmd
@@ -45,13 +45,15 @@ Lets take a look at this data.
 class(dat)
 ```
 
-R has loaded the contents of the .csv file into a variable called `dat` which is a `data frame`.
+R has loaded the contents of the .csv file into a variable called `dat` which is a `data frame`. A data frame is a list of variables of the same number of rows with unique row names, given class "data.frame".
 
-```{r dimDat}
-dim(dat)
+We can compactly display the internal structure of a data frame using the  structure function `str`.
+
+```{r strDat}
+str(dat)
 ```
 
-The data has 100 rows and 9 columns.
+The `str` function tell us that the data has 100 rows and 9 columns. It is also tell us that the data frame is made up of character `chr`, integer `int` and `numeric` vectors.
 
 ```{r headDat}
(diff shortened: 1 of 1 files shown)


katrinleinweber:
Thanks for these suggestions :-) We'll review them during the weekend.

_episodes_rmd/10-supp-addressing-data.Rmd
```diff
@@ -45,13 +45,15 @@ Lets take a look at this data.
 class(dat)
 ```
 
-R has loaded the contents of the .csv file into a variable called `dat` which is a `data frame`.
+R has loaded the contents of the .csv file into a variable called `dat` which is a `data frame`. A data frame is a list of variables of the same number of rows with unique row names, given class "data.frame".
```
katrinleinweber: Is a `the` or similar missing in the end?

_episodes_rmd/10-supp-addressing-data.Rmd
```diff
@@ -147,6 +149,13 @@ We usually use the `$` operator to address a column by name
 dat$Gender
 ```
 
+When we address a data frame by a single column, R will return a vector of that column class and not a data frame.
```
katrinleinweber: I don't think "address a data frame by a single column" is a good description of the `$` operation. Please rephrase, or leave it to the explanation just above the `dat$Gender` code block.

_episodes_rmd/10-supp-addressing-data.Rmd
```diff
@@ -45,13 +45,15 @@ Lets take a look at this data.
 class(dat)
 ```
 
-R has loaded the contents of the .csv file into a variable called `dat` which is a `data frame`.
+R has loaded the contents of the .csv file into a variable called `dat` which is a `data frame`. A data frame is a list of variables of the same number of rows with unique row names, given class "data.frame".
 
-```{r dimDat}
-dim(dat)
```
katrinleinweber: I agree with this removal, because we already explain it in https://github.com/swcarpentry/r-novice-inflammation/blob/df5f83988a0bf12da7cf4713742a4ac63e0686de/_episodes_rmd/01-starting-with-data.Rmd#L296-L300

_episodes_rmd/10-supp-addressing-data.Rmd
katrinleinweber: Also, let's try to not repeat "data frame" or "data.frame".

_episodes_rmd/10-supp-addressing-data.Rmd
ggrimes: Would "when we subset the data frame by a single column" be a better alternative?

_episodes_rmd/10-supp-addressing-data.Rmd
ggrimes: The sentence " A data frame is a list of variables of the same number of rows with unique row names, given class "data.frame" was taken from the R help page on data frames (https://stat.ethz.ch/R-manual/R-devel/library/base/html/data.frame.html). I am happy to remove it from the pull request if it is not helpful or unclear.

_episodes_rmd/10-supp-addressing-data.Rmd
ggrimes: Or "When we extract a single column from a data frame"?

katrinleinweber: The `extract`-variant is better IMHO.

`class(subset(dat, select = BloodPressure))` on the other hand would still be a `data.frame` and thus wouldn't fit to the 2nd part of your sentence.

_episodes_rmd/10-supp-addressing-data.Rmd
katrinleinweber: Yes, in that case, please remove it. We don't need to duplicate content from the official manual. On the contrary, it seems to require some [improvement suggestions](https://www.r-project.org/bugs.html) itself ;-)

katrinleinweber:
Cool, thank you @ggrimes :-)

PR #458, opened 2019-11-05
Updated lesson 2 exercise to include all solutions

SoftwareMonk:
The exercise `Named Variables and the Scope of Variables` only had a solution for one of the two questions. I added the answer to the first question, not sure if we want to expand on it more? I'm open to suggestions.

P.s. this is not part of checkout, I'm already certified 😃


HaoZeke:
Thanks for this (even after checkout :D)! I personally feel like this is more than enough detail. :)
The instructors taking the workshop can assign more time if they need to I guess.

Issue #343, opened 2018-03-27
Packaging: refer to other resources


katrinleinweber:
https://github.com/swcarpentry/r-novice-inflammation/_episodes_rmd/08-making-packages-R.Rmd could use a mention of https://devguide.ropensci.org/building.html (formerly [onboarding.../packaging_guide](https://web.archive.org/web/20171222124034/http://onboarding.ropensci.org/packaging_guide.html)).

katrinleinweber:
Related to #360, because packaging is touched there as well.

lauzikaite:
The suggested link [](https://onboarding.ropensci.org/packaging_guide.html) is no longer valid.

@katrinleinweber, did you mean the rOpenSci book that is now live at [](https://devguide.ropensci.org/)? If so, I can submit a PR shortly.

katrinleinweber:
It's probably https://devguide.ropensci.org/building.html now, at least when compared to https://web.archive.org/web/20171222124034/http://onboarding.ropensci.org/packaging_guide.html

lakikowolfe:
Hey! I was just going through issues to find one to work on as a part of my instructor training. I believe this issue has been resolved in #459 and should be marked as closed.

HaoZeke:
Thanks @lakikowolfe !

PR #459, opened 2019-11-11
adding a link to rOpenSci guidelines to episode 8

lauzikaite:
Adding a link to rOpenSci guidelines book to episode 8, as suggested in issue #343


HaoZeke:
Thank you for taking the time to contribute. Perhaps a little rewording could make this clearer. The rOpenSci are not necessarily optimal, they are simply the requirements of that community. Especially since a lot of their guidelines will necessarily be out of scope for many people finishing this workshop (e.g. the CI guidelines) I am not sure how this should be best phrased.

katrinleinweber:
How about swapping this with the `"R packages" book`? Like:

> There are other optional components: [rOpenSci][r-open-sci] has written a science-focused guidebook, while [the "R packages" book][r-pkgs-desc] has all the fundamental information.

Issue #452, opened 2019-10-15
Setting directories with tab auto-complete


michaelofrancis:
Hi all,

In the first lesson ("Analyzing Patient Data"), when an absolute beginner is setting their working directory to load in csv files, it may be useful to let them know that they can use the tab key within the setwd("") quotes to see a list of possible directories. I imagine that getting some folks to type in their working directory correctly may be difficult in a live session and this could make it easier.

diyadas:
Thanks for the suggestion. I do have some concern about growing the size of this episode. Perhaps it’s possible to reference the Unix shell on tab completion (as it’s frequently taught before this lesson) and this keep the reference short?

Thoughts @katrinleinweber ?

katrinleinweber:
When I +1'ed the above, I assumed that we're talking only about 2 or 3 sentences, maybe scattered throughout 1 or 2 episodes to nudge learners towards the `tab`. A reference to https://swcarpentry.github.io/shell-novice/02-filedir/index.html#nelles-pipeline-organizing-files is a good idea in any case :-)

diyadas:
Cool! If anybody wants to submit a PR, happy to review.

PR #465, opened 2020-02-05
Set directories using tab completion

ryantaylor-cu:
This is an attempt to submit a PR for https://github.com/swcarpentry/r-novice-inflammation/issues/452

As suggested in this issue, I added a couple sentences around using tab completion.

I only referred the Unix Shell in general.  Would it be better to include a link to the relevant place in the Shell Novice lesson? 

I also changed the pronoun in the following sentence to match ('we' instead of 'you').  Is that OK to include in the same PR?


HaoZeke:
Thanks for this! I believe all the changes are pertinent and well within the scope of a single change. 👍

Issue #466, opened 2020-02-11
Episode 1: "Printing with Parentheses"


kekoziar:
If we're modeling proper data management and reproducible science, the example in "Printing with Parentheses"  `(total_weight <- weight_kg + weight_lb)` should have one of the variables changed so the units are the same.  perhaps `(total_weight <- weight_kg*2.2 + weight_lb)`

HaoZeke:
Excellent point. Thanks! Would you be willing to open a PR resolving this?

kekoziar:
Yes, @HaoZeke.  Thanks!

EastBayEv:
Awesome trick!

HaoZeke:
I'll be closing this now, thanks again!!

PR #467, opened 2020-02-12
Ep 1: Callout "Printing with Parentheses"

kekoziar:
Per issue #466, change computation in "Printing with Parentheses" so the units are the same.


HaoZeke:
Excellent 🎊

Issue #470, opened 2020-04-20
addressing data by row names too in 10. Addressing data


sarahbonnin:
Hello,

In the supplementary section "10. addressing data", I would like to suggest a small change.

At the very beginning, it is written:
"
There are three main ways for addressing data inside R objects.
    By index (subsetting)
    By logical vector
    By name (columns only)
"

The "By name (columns only)" gives the impression that data can't be accessed by row names, which isn't true (even though probably much less used than fetching columns by names).

If you think it is relevant to add this information, here is my suggestion on how to add it in a few lines:

```{r}
# check row names for the data that was already imported
rownames(dat)

# re-read data file (save in a new variable dat2), using the first column (ID) as row names:
dat2 <- read.csv(file = 'data/sample.csv', header = TRUE, stringsAsFactors = FALSE, row.names=1)

# check row names:
rownames(dat2)

# show that you can extract the data for a specific row, using its name:
dat2["Sub072", ]

# or several rows at once, using the "combine" function:
dat2[c("Sub009", "Sub072"), ]
```

Best,
Sarah

HaoZeke:
Hi Sarah, thanks for this suggestion. Would you like to open a PR for this? I agree that this is probably a good idea to mention.

sarahbonnin:
Hello,
Sure, I will do this!

HaoZeke:
Thanks so much for the follow up, and the excellent PR. ^_^

PR #471, opened 2020-04-27
add addressing data by row names

sarahbonnin:
As commented in issue [#470](https://github.com/swcarpentry/r-novice-inflammation/issues/470), I suggest to add a small section in the "addressing data" lesson to address data by row names.


HaoZeke:
Fantastic, thanks so much ^_^

Issue #474, opened 2020-05-20
Fluid Representations in Addressing Data


eigensong:
Dear Carpentries Community,

In my instructor training I learnt about "fluid representations" where multiple words for the same concept can confuse learners. I noticed that in the [Addressing Data](http://swcarpentry.github.io/r-novice-inflammation/10-supp-addressing-data/index.html) most of the lesson is really careful to avoid fluid representations, by consistently using the word addressing to describe accessing parts of a data frame, rather than mixing in words like subsetting or indexing. However, the titles for the three sections of the types of addressing are:
Addressing by Index
Addressing by Name
Logical Indexing

While I agree that "logical indexing" is the standard way to call this, and that students should be made aware of it, I think it might be clearer to change the title to "Addressing by Logical Vector" (or something similar) and then mention in that section that this is often referred to as "logical indexing." 

Any thoughts?

Cheers,

Elena

HaoZeke:
Hi Elena,

That's a really great point regarding the text consistency. I think as long as we drop the term Logical Indexing in the main text we should be good to go with replacing the heading. Would you be willing to open a pull request with this change?

Thanks a ton,
Rohit

eigensong:
Hi Rohit,

Thanks for your feedback. I'm happy to try open a pull request with this change, but I'm still learning how to use git, so it might take me a little while to work my way through the instructions.

Cheers,

Elena

HaoZeke:
Hi Elena,

Thanks for the follow up! Please take all the time you need, and thanks again.

---
Rohit

eigensong:
Hi Rohit,

I've opened this pull request https://github.com/eigensong/r-novice-inflammation/pull/1 . Should I have somehow linked it to this open issue?

Cheers,

Elena

HaoZeke:
Hi Elena,
Thanks for the PR! It links back to this issue so I can assign it to be closed once we merge the PR in :)
---
Rohit

eigensong:
Hi Rohit,

I tried another pull request. It's linked above. Let me know if I need to do anything else.

Cheers,

Elena

HaoZeke:
Thanks @eigensong!

PR #476, opened 2020-06-01
Fix logical indexing terminology.

eigensong:
Dear Carpentries Community,

In my instructor training I learnt about "fluid representations" where multiple words for the same concept can confuse learners. I noticed that in the [Addressing Data](http://swcarpentry.github.io/r-novice-inflammation/10-supp-addressing-data/index.html) most of the lesson is really careful to avoid fluid representations, by consistently using the word addressing to describe accessing parts of a data frame, rather than mixing in words like subsetting or indexing. However, the titles for the three sections of the types of addressing are:
Addressing by Index
Addressing by Name
Logical Indexing

I raised this in Issue https://github.com/swcarpentry/r-novice-inflammation/issues/474

Here I change the last sections to use addressing instead of indexing in line with the rest of the document and introduce the term logical indexing as an alternative.

I hope that helps,

Elena

### DIFF
--- a/_episodes_rmd/10-supp-addressing-data.Rmd
+++ b/_episodes_rmd/10-supp-addressing-data.Rmd
@@ -190,7 +190,7 @@ We can now extract one or more rows using those row names:
 dat2["Sub072", ]
 ```
 
-```{r_row_names_addressing_2}
+```{r row_names_addressing_2}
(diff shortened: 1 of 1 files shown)


HaoZeke:
Excellent, thanks!

Issue #481, opened 2020-06-29
Style guide, tiny fix, and error handling in function writing


taddallas:
Analyzing patient data::Variable Naming Conventions: mention the Google R style guide here in terms of variable and function naming? https://google.github.io/styleguide/Rguide.html

Creating Functions::Create a Function work problem: `edges` function has odd syntax in indent from first line to second. This is an incredibly tiny fix. Related to introducing functions, is it worth discussing error handling (perhaps around where testing and documentation are discussed)? An example would be if the `center` function were handed an NA or a character string. How to error out gracefully and informatively seems important to function writing, but maybe this is too much for this section?

Happy to contribute to these fixes after the main devs weigh in on whether or not they should be incorporated. 

((also, a side note: there are a lot of outstanding open issues which have already been had successful PRs and should probably be closed. I'm a new instructor looking to contribute -- partially to fulfill the requirement of contributing -- but kept finding open issues that were already discussed and addressed))

HaoZeke:
Hi @taddallas, sorry for the long silence! With regards to style guides, I prefer the original [tidyverse one](https://style.tidyverse.org/) but maybe a PR with links to both would be a good idea.

I think a tasteful introduction to error handling in the text would be a great idea, since in practice many new learners do often end up with errors.

It'd be great to have these in a PR (or series of PRs possibly).

((re: side note: I took over a while back, it'd be great if you could just leave a message on the issues you find which are redundant, that way I could eliminate them quicker ^_^ ))

Thanks again!

taddallas:
Hi @HaoZeke, no worries about the silence. You're right that the difference is minimal, and the tidyverse style guide link you provided makes it seem like there is lots of overlap, and maybe even some feedback and convergence between the Google style guide and the tidyverse style guide. 

I'll try to push along a short introduction to error handling in the assignment shortly and will submit a PR soon-ish. 

I'll go through and flag issues that I believe are resolved, but this may be a bit subjective. I'll focus on those that have merged PRs, as these are the most likely to have been addressed.

taddallas:
Realized a bit late, but functions are introduced before conditionals in the current setup, so error handling is a bit tough to introduce here (as handling errors is much easier to discuss after learning `if` statements). Submitted a PR with my edits, which simply introduce error handling as an important concept.

taddallas:
Closed for now. Can be reopened if the flow of the lesson changes to introduce conditionals before function writing, as this flow would better address error handling alongside of function writing.

PR #482, opened 2020-07-18
weak introduction to error handling

taddallas:
Attempt to address issue #481 , adding in a bit on error handling. Might flow better if conditionals were introduced before function writing. It reads as if R is a functional programming language.


HaoZeke:
Thanks for this! We might update the flow later (am in touch with the lesson developers) and I'll reach out to you again for inputs when that progresses further.

PR #488, opened 2020-10-12
remove-iris

ErinBecker:
Replaces example using `iris` dataset with example using `mtcars` dataset to avoid using data created by a eugenicist.


Alanocallaghan:
The data itself was gathered by Edgar Anderson (not a eugenicist as far as I can tell), but given the heavy association with RA Fischer it's understandable to want to remove, and it's not like it adds any particular value above other examples.

PR #492, opened 2020-10-27
Strings as factors default

phadenz:
Suggested update of Episode 11: Reading and Writing CSV Files from Software Carpentry lesson Programming with R.

The current version of this episode contains an example that assumes the default value for argument read.csv(stringsAsFactors) is TRUE. From R version 4, the default for stringsAsFactors has changed to FALSE (cf. https://developer.r-project.org/Blog/public/2020/02/16/stringsasfactors/), so the example code does not behave as described. The suggested edits modify the example to match the new default and add a little additional discussion of the advantages/disadvantages of factor data columns.


jeffjshi:
This would be an excellent way to discuss and/or to introduce a best-practice of including default arguments in your scripts and code, especially if they are absolutely integral to your analyses/output, since they do change over time and over R versions. This is maybe even doubly important for those researchers who may depend on old, unmaintained R packages that need to run on older R versions!

Perhaps a simple exercise is in order that demonstrates how the same data.frame can result in drastically different output to downstream functions depending on the stringsAsFactors argument - even simple functions like str(). This could further lead to a general programming lesson about how -not- seeing an error doesn't necessarily mean your code is doing what you want to do! While this doesn't necessarily align with objectives for this episode (although it is a deeper dive into "explore the arguments of the csv reader"), it ties directly into later ones and their objectives (e.g. "Understanding Factors" and "Data Types and Structures").

PR #497, opened 2021-01-27
Fix block quote formatting

zkamvar:
There was a gap in the block quote, which is find for Kramdown, but not so fine for commonmark or github's markdown parser (see the same section here: https://github.com/swcarpentry/r-novice-inflammation/blob/6108ab24cee6fece6671f97ed6b8f4cfe8552ded/_episodes/11-supp-read-write-csv.md).


HaoZeke:
Excellent, thanks!

PR #499, opened 2021-02-10
R version 4 fix for change in stringsAsFactors behaviour

NatePollard:
From R version 4.0 the default behaviour for loading csv files has changed. This has broken episodes 11 and 12 in the course. 

 `stringsAsFactors = TRUE` was the default behavior for R prior to version 4.0. From version 4.0 this has changed to  `stringsAsFactors = FALSE`.

This fix makes some code and text changes to fix episodes 11 and 12.


HaoZeke (approved):
Excellent, thanks for the comprehensive changes, @NatePollard

PR #500, opened 2021-02-10
StringAsFactor is no longer TRUE by defaul

shmuhammad2004:
We need to update the lesson 12 in totality since StringAsFactors is no longer = TRUE in R by default. [This was announced at the R Core meetings in Toulouse in 2019 ](https://developer.r-project.org/Blog/public/2020/02/16/stringsasfactors/).

Also, from the [setup page](https://swcarpentry.github.io/r-novice-inflammation/setup.html), the lesson categorically said, "This lesson assumes you have current versions of the following installed on your computer". Therefore, the current version of R Software default to  StringsAsfactors = FALSE.

### DIFF
--- a/_episodes_rmd/12-supp-factors.Rmd
+++ b/_episodes_rmd/12-supp-factors.Rmd
@@ -21,6 +21,12 @@ source('../bin/chunk-options.R')
 knitr_fig_path("12-supp-factors-")
 ```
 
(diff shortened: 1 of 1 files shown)


HaoZeke:
Hi @shmuhammad2004, thanks so much for this, but the changes are a little outdated compared to the current main, since #499 was merged. Would you be willing to open a new PR? Perhaps with some more clarifications around `stringsasfactor`? Thanks again for this!

PR #501, opened 2021-02-11
remove iris from exercise

ErinBecker:
Follow up from https://github.com/swcarpentry/r-novice-inflammation/pull/488. In this case, using `PlantGrowth` dataset as it illustrates the point the exercise is trying to make that dataframes can have elements of different types.


HaoZeke (approved):
Looks alright, and seems like this change is the will of the community! Thanks a ton @ErinBecker.

PR #502, opened 2021-02-11
Fix ruby build action in GitHub workflows

NatePollard:
Fix to the ruby build action which is causing errors in the GitHub workflow continuous integration actions.


HaoZeke:
Excellent, thanks!

PR #505, opened 2021-03-15
Update styles

zkamvar:
Note for future style updaters: Conflict in bin/boilerplate/CONTRIBUTING.md due to b50bada in #418 

There was also a quick fix applied to the GitHub action that I updated to the current styles.

I also removed "deploy.yml" because the "website.yml" action now takes care of this process.


HaoZeke:
Thanks!

zkamvar:
@HaoZeke, one thing I did not take into account: this assumes that we have changed the default branch to "main" and not "master". 

If you would like, I can do that for you, but it does mean that some manual work needs to be done for your clone of the repo:

```
git branch -m master main
git fetch origin
git branch -u origin/main main
```

Would you like me to migrate this from master to main?

HaoZeke:
Sure, I can update my clone. I'm a bit concerned about the 344 forks but most don't really track this anyway. Thanks!

PR #510, opened 2021-04-09
Update alt text for R package directory image

### LINKED ISSUE #372: Make alt text for images less vague?
The alt text for some lesson images does not specifically describe what is shown in the image. While this may not be as necessary for “supplemental” images that reinforce ideas already described in the text, some of the writing implies that the user must see/understand what the image conveys.

A good example of this pattern is found in the [Making R Packages section](http://swcarpentry.github.io/r-novice-inflammation/08-making-packages-R/) where the text reads: “Overall, your package directory should look something like this:” and the alt text for the corresponding image says, “R Package Structure.”

LoraArmstrong:
Fixes #372 

More descriptive alt text for the R package directory image, which contains material that is not described in the main text of the lesson.

### DIFF
--- a/_episodes_rmd/08-making-packages-R.Rmd
+++ b/_episodes_rmd/08-making-packages-R.Rmd
@@ -147,7 +147,7 @@ The /man directory has a .Rd file for each .R file with properly formatted docum
 
 Overall, your package directory should look something like this:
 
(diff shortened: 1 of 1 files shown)


HaoZeke:
Wonderful! Thanks so much @LoraArmstrong

PR #514, opened 2021-04-14
Add note about tests with NAs and is.na()

### LINKED ISSUE #511: Add a section about NA values in tests
In the [Making Choices](https://swcarpentry.github.io/r-novice-inflammation/04-cond/index.html), I think it would be extremely valuable to add a section / a note / a box about how to test for the presence of `NA`s with `is.na()`:

- It is a very common source of confusion and errors for beginners and the error message is not easy to understand:

``` r
a <- NA 

a == NA
#> [1] NA

if (a == NA) {
  print("Hi!")
}
#> Error in if (a == NA) {: missing value where TRUE/FALSE needed
```

<sup>Created on 2021-04-13 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0.9002)</sup>

- A similar note is already present about testing for `NULL` values with `is.null()`

- `NA` values have already been introduced in the previous lesson about `for` loops

If you think this would be a worthwhile addition, I'm happy to submit a PR for it.

Bisaloo:
Fix #511.

I'm happy to revisit this PR as much as needed. Please let me know if you see any possible improvements.

### DIFF
--- a/_episodes_rmd/04-cond.Rmd
+++ b/_episodes_rmd/04-cond.Rmd
@@ -199,6 +199,33 @@ if (1 > 0 || -1 > 0) {
 
 In this case, "either" means "either or both", not "either one or the other but not both".
 
+> ## Special case of the `NA` variable
+> 
+> You may remember from the previous lesson that R has a special variable, `NA`,
+> for designating missing values. Because it represents the absence of a value,
+> you will not be able to test its equality or inequality with another value.
+> Such tests always return `NA`:
+> ```{r test-equal-NA}
+> a <- NA
+> a == 1
+> a == NA
+> ```
+> You need to be particularly careful in tests because trying to compare a `NA`
+> value will result in an error:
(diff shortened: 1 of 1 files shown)


HaoZeke:
Hi @Bisaloo, thanks so much for this! Looks great, and the examples are comprehensive and clear. However, I was wondering if you'd be willing to change from "You" to "We", for consistency with the rest of the lesson.

Bisaloo:
Hi @HaoZeke, thanks for the feedback. It makes sense and I've updated this PR accordingly. There is still one 'you' at the very beginning because it's an actual 'you' in this case. Hope that's okay.

Please let me know if you see something else.

HaoZeke (approved):
Fantastic! Looks good to me (LGTM)!

HaoZeke:
Thanks for your contribution @Bisaloo!!

Issue #512, opened 2021-04-13
setwd in beginning of lesson is bad practice


Alanocallaghan:
Running setwd in the beginning of an R session is horrible practice, but it's currently part of the [first lesson](http://swcarpentry.github.io/r-novice-inflammation/01-starting-with-data/index.html).

This should be changed, if for no other reason than for fear of having [Jenny Bryan come and set learners' computers on fire](https://www.tidyverse.org/blog/2017/12/workflow-vs-script/).

Also this is detailed as bad practice [later](http://swcarpentry.github.io/r-novice-inflammation/06-best-practices-R/index.html).

I can see two ways of doing this:

1. include an RProject in the zip download
2. tell users to start by making a project.

I would favour the latter personally.

Actually reading through the lesson - episode 6 says "don't setwd" but episode 11 is back at it again.

If I'm being pedantic (apparently I am) I would also say:
- "nested functions" [here](http://swcarpentry.github.io/r-novice-inflammation/02-func-R/index.html) are nested calls not nested functions.
- `stopifnot(action %in% c("--min", "--mean", "--max"))` should probably be  
  `match.arg(action, choices = c("--min", "--mean", "--max"))` [here](http://swcarpentry.github.io/r-novice-inflammation/05-cmdline/index.html),
  or at least `match.arg` should be mentioned as an alternative.

I'm happy to make a PR to change these if maintainers think they're not awful ideas

HaoZeke:
I think these are all great! Would love to see a PR with these. Thanks :)

PR #515, opened 2021-04-14
Use projects not setwd

Alanocallaghan:
Address #512


Alanocallaghan:
I realise I've done this a bit backwards and the project setup should prob be in setup.md. Oops!

_episodes_rmd/01-starting-with-data.Rmd
```diff
@@ -54,19 +56,38 @@ We want to:
 
 To do all that, we'll have to learn a little bit about programming.
 
-### Loading Data
+# Getting started
```
HaoZeke, line 59: Perhaps this should be ## here?
```suggestion
## Getting started
```

Otherwise the lower headings need to go up one level as well...

_episodes_rmd/01-starting-with-data.Rmd
```diff
@@ -54,19 +56,38 @@ We want to:
...
+  not there already.
 
-Just like in the Unix Shell, we type the command and then press <kbd>Return</kbd> (or <kbd>Enter</kbd>).
-Also similar to the Unix Shell, we can use tab completion to make it easier to type the working directory. So while typing inside the setwd command's quotes, we can press <kbd>Tab</kbd> to list possible directories. 
-Alternatively we can change the working directory using the RStudio GUI using the menu option `Session` -> `Set Working Directory` -> `Choose Directory...`
 
-The data file is located in the directory `data` inside the working directory. Now we can load the data into R using `read.csv`:
+### Loading Data
+
+Now that we are set up with an Rstudio project, we can be sure the data
```
HaoZeke, line 87:
```suggestion
Now that we are set up with an Rstudio project, we are sure that the
```

HaoZeke (changes requested):
Looks great to me! Thanks for this. I made some minor suggestions, but otherwise this is fantastic.

HaoZeke (approved):
Thanks for the changes, looks great!

Alanocallaghan:
Thanks!

PR #520, opened 2021-05-26
Update 11-supp-read-write-csv.Rmd

piperfw:
Since `ifelse` strips attributes, if `str(carSpeeds)` (now line 100) is called after line 82, the output has `$Color` as `chr` not `Factor`. This is incongruent with the point being made.


HaoZeke (approved):
Excellent, great catch.

Issue #525, opened 2021-09-06
Lesson contribution 2. Creating Functions


wfpinzonp:
Dear SWC team

I found one small improvement that can be made to the Programming with R. Particularly in the second section "Creating Functions" (http://swcarpentry.github.io/r-novice-inflammation/02-func-R/index.html).

Very little introduction to the building of the function is given, I believe a general explanation and structure to the function can be provided right below the title **Defining a Function** as:

You can write your own functions in order to make repetitive operations using a single command. Let's start by defining your function "my_function" and the input parameter(s) that the user will feed to the function. Afterwards you will define the operation that you desire to program in the body of the function within curly braces ({}). Finally, you need to assign the result (or output) of your function in the return statement.

```
my_function <- function(input) {
# perform action and produce output
return(output) # return output value
}
```

I really hope this helps. I'm attentive to any feedback/changes that you might have.

Best regards

William **Pinzon Perez**

HaoZeke:
Hi @wfpinzonp. Thanks for this! Would you be willing/able to open a PR (pull request) with this contribution? Let me know if you need help setting that up.

wfpinzonp:
Hi Rohit

Thank you for your email. I'd really appreciate if you can guide me through the process of opening a PR. I tried but got a bit confused.

Kind Regards


William Pinzon-Perez

Master of Applied Econometrics

The University of Queensland

Email: ***@***.***

Phone: 0431 628 820

________________________________

PR #526, opened 2021-09-08
Include explanatory text

wfpinzonp:
I suggest that we explain the syntax of a function before we go into the first example. This is following up from issue #525: https://github.com/swcarpentry/r-novice-inflammation/issues/525


HaoZeke (approved):
Wonderful, thanks so much.

HaoZeke:
Closes #525

PR #528, opened 2022-01-17
off by one error fixed

steltenpower:
'every second something' doesn't start with the 'first something', does it?


steltenpower:
Nothing to see here.
Was an oops.
The error was a little different and already fixed and committed

PR #529, opened 2022-01-17
This time correctly I think: off-by-1 error fixed

steltenpower:
'every second day' should start with day 2, right?
Minutes ago I sent in a similar thing, but that was wrong.


HaoZeke:
LGTM, thanks!

Issue #521, opened 2021-06-09
Typo (missing word?) in Analyzing Patient Data


sraorao:
In the Loading Data section, http://swcarpentry.github.io/r-novice-inflammation/01-starting-with-data/index.html

> we are sure that the and scripts we are using are all in our working directory

Should that be _...the **data** and scripts..._?

zkamvar:
fixed in #534. Thank you for opening this issue!

Issue #495, opened 2020-12-14
`c` function means concatenate, not combine


reblake:
I'm currently teaching this SWC R lesson, and I noticed an incorrect definition of the `c` function in the [01-starting-with-data section](http://swcarpentry.github.io/r-novice-inflammation/01-starting-with-data/index.html).    It says the `c` function means "combine", but it actually means "concatenate".   Correcting this definition will improve clarity for learners.

Jariati:
I learnt from some R resources. I think 'c' function can be both, concatenate or combine. There is an explanation: if factor concatenation by c() should give a factor. It means, when user use c() function, they combine some details. In other words, I don't think by replacing combine with concatenate will provide significant changes in clarity.

HaoZeke:
We would like to stick to the standard definition from `?c`:
![image](https://user-images.githubusercontent.com/4336207/103251983-f3c7f780-4972-11eb-9804-5199bd9f09f7.png)

reblake:
@HaoZeke, the definition of the function is not the name of the function.  The _name_ of the function is  "concatenate" ("c" for short), while the _definition_ of the function is what it does ("combine values...").   
See the comments here: https://stackoverflow.com/questions/11488820/why-use-c-to-define-vector
Also, see the R manual section 2.1: https://cran.r-project.org/doc/manuals/r-release/R-intro.pdf

Many functions can combine objects in R, so being specific that this functions is concatenate (vs. paste, cat, etc.) is important for learners, especially those new to R.

HaoZeke:
@reblake my apologies. For posterity I'm also putting the relevant section of the manual (linked)

![image](https://user-images.githubusercontent.com/4336207/103582701-4dca3f00-4ed6-11eb-9f30-83bb7f06af6f.png)

However, it seems to still be a little contested, not an authoritative reference at all but [R For Dummies](https://www.dummies.com/programming/r/how-to-combine-vectors-in-r/) feels happy to call it `combine`.

![image](https://user-images.githubusercontent.com/4336207/103583052-f1b3ea80-4ed6-11eb-98d1-a8eac665c4d8.png)

Further muddying the waters, I normally teach `paste` for concatenate, simply because of the R definition.

![image](https://user-images.githubusercontent.com/4336207/103583141-1e680200-4ed7-11eb-9a6a-1aa8d28d8ff8.png)

[...]

PR #537, opened 2022-03-11
Typo fix & changing c() description

lhami:
Fixing typo from #521 and changing description of `c()` as suggested in #495 

The use of "**c**ombine" as a mnemonic is used in other parts of the lesson, at the very least episode 02 (and I think having some meaningful word to tie the one-letter function to is helpful for novices), so I kept it while changing the wording so it doesn't imply the function is named one or the other.


zkamvar:

Thank you for the changes @lhami! I agree that this definitely makes the wording around the `c()` function _much_ clearer for people new to R.

PR #547, opened 2022-07-06
Update 05-cmdline.Rmd

cknuepfer:
Correct output of print-args.R

Running `Rscript print-args.R` returns `--no-echo`instead of `--slave`.

### DIFF
--- a/_episodes_rmd/05-cmdline.Rmd
+++ b/_episodes_rmd/05-cmdline.Rmd
@@ -116,7 +116,7 @@ From the R help file:
 Thus running a file with Rscript is an easier way to run the following:
 
 ```{r engine='bash'}
-R --slave --no-restore --file=print-args.R --args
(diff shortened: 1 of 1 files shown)


cknuepfer:
There are two more mentions of `--slave` in this lesson which should also be replaced.

I think the option `--slave` is from an older R version and now deprecated because of it is dismissive language.

zkamvar:

Considering that this change of parameter is needed, [is now over 3 years old](https://stat.ethz.ch/pipermail/r-help/2019-September/464273.html), and precedes the update to R version 4.0, this text change is very welcome. Because you mention it is present in other places in the episode, I will edit this PR to make the subsequent changes.

zkamvar:
The checks are failing through no fault of your own, but I am going to merge this because I am confident that it will pass.

PR #548, opened 2022-07-06
Remove not yet introduced stdin

cknuepfer:
In the episode this solution is shown **before** the handling of the standard input. It is confusing to have `file("stdin")` at this point.


zkamvar:

Thank you for the contribution @cknuepfer! Looking at where this is in the lesson and the fact that it is only used once, I agree with your assessment and will merge this.

PR #554, opened 2022-10-12
add alt text

ErinBecker:
Add alt text written by McKenzie Hagen for https://swcarpentry.github.io/r-novice-inflammation//fig/rmd-03-loops-R-inflammation-01-1.png.


zkamvar:
fixed in #571 due to multiple plots created by the same chunk

PR #555, opened 2022-10-12
update alt text

brynnelliott:
Added alt text by McKenzie Hagen for https://swcarpentry.github.io/r-novice-inflammation//fig/rmd-03-loops-R-inflammation-02-1.png

PR #560, opened 2022-10-12
added alt tex

brynnelliott:
Alt text written by McKenzie Hagen for https://swcarpentry.github.io/r-novice-inflammation//fig/rmd-03-loops-R-inflammation-02-3.png

### DIFF
--- a/_episodes_rmd/03-loops-R.Rmd
+++ b/_episodes_rmd/03-loops-R.Rmd
@@ -45,7 +45,7 @@ analyze("data/inflammation-01.csv")
 
 We can use it to analyze other data sets one by one:
 
-```{r inflammation-02}
+```{r inflammation-02, fig. alt = "A scatterplot of minimum daily inflammation over 40 days, increasing in a step-wise pattern until day 20, and then decreasing in a step-wise pattern to visualize inflammation levels in a single dataset."}
 analyze("data/inflammation-02.csv")
 ```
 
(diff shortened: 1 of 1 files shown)


_episodes_rmd/03-loops-R.Rmd
```diff
@@ -45,7 +45,7 @@ analyze("data/inflammation-01.csv")
 
 We can use it to analyze other data sets one by one:
 
-```{r inflammation-02}
+```{r inflammation-02, fig. alt = "A scatterplot of minimum daily inflammation over 40 days, increasing in a step-wise pattern until day 20, and then decreasing in a step-wise pattern to visualize inflammation levels in a single dataset."}
```
brynnelliott, line 48:
```suggestion
```{r inflammation-02, fig.alt = "A scatterplot of minimum daily inflammation over 40 days, increasing in a step-wise pattern until day 20, and then decreasing in a step-wise pattern to visualize inflammation levels in a single dataset."}
```

zkamvar:
This is a duplicate of #555

zkamvar:
Note: this is NOT a duplicate of #555, but an extension. In [knitr 1.31](https://github.com/yihui/knitr/releases/tag/v1.31), it is possible to include alt text for images (emphasis mine):

> Added a new chunk option fig.alt for users to specify the alternative text in the alt attribute of the  tag (images). Previously, the alt attribute takes value from the chunk option fig.cap (it still does so if fig.alt is NULL). **If there are multiple plots/images in a chunk, you can pass a character vector to fig.alt, and it will be applied to individual images** (thanks, [@]cderv, https://github.com/yihui/knitr/pull/1900).

So, the solution is a bit awkward but what you can do is this (replacing the placeholder text for the actual alt text that describes the image:

```{r}
{r, plot-name, fig.alt = c("... alt text 1...", "... alt text 2...", "... alt text 3...")}
```

PR #567, opened 2022-10-27
added alt text

brynnelliott:
Added alt text by McKenzie Hagen for https://swcarpentry.github.io/r-novice-inflammation//fig/rmd-03-loops-R-loop-analyze-4.png


zkamvar:
fixed in https://github.com/swcarpentry/r-novice-inflammation/pull/571 due to multiple plots created by the same chunk

PR #574, opened 2023-02-17
Replace link to R Packages book style chapter with one to the tidyverse style guide

alburycatalina:
Hey there! I noticed that the link in Episode 1 (Analyzing Patient Data) to the style chapter in the R Packages book was broken. 

The [2nd edition of R packages](https://r-pkgs.org/code.html#code-style) refers readers to the tidyverse style guide, so perhaps the lesson should instead bring the learner to the ["Object Names" section of style guide](https://style.tidyverse.org/syntax.html) instead.


zkamvar:
Thank you!

PR #591, opened 2023-08-30
Update 01-starting-with-data.Rmd

annajiat:
Earlier the participants were told to create script.R but here the lesson expects them to run by only pressing ENTER which does not work in the script.R file. We have to type CTRL+ENTER. Added changes to accommodate for both scenarios.

### DIFF
--- a/episodes/01-starting-with-data.Rmd
+++ b/episodes/01-starting-with-data.Rmd
@@ -180,7 +180,7 @@ Once a variable is created, we can use the variable name to refer to the value i
 
 <img src="fig/tag-variables.svg" alt="Variables as Tags" />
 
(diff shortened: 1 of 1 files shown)


annajiat:
Working in the Script.R is recommended as it is helpful for future reproducibility.

Bisaloo:
Thanks for the catch! On my own machine at least, both 'Enter' and 'CTRL+Enter' work in the console so maybe we should mention just this solution since it covers all situation.

What do you think?

HaoZeke (approved):

HaoZeke:
Thanks @annajiat and @Bisaloo. I went ahead and put this in, in these cases more is better (since it doesn't add too many lines anyway).

PR #598, opened 2023-11-11
Replace excel screenshots by embedded file contents

### LINKED ISSUE #434: CSV file photos are screen shots of Excel files instead of CSV files
Seems like a great learning tool, but I am curious why do the photos of supposed CSV file outputs have photos of EXCEL spreadsheets instead of the actual CSV files?  CSV files are just plain text files.   Showing them as if they were EXCEL files will confuse novice programmers into thinking that they are spreadsheets. Plus if you open a CSV file in EXCEL then there is extremely large probability that EXCEL will not properly display the content.

Bisaloo:
Fix #434.

This will also be hopefully more accessible for visitors using screenreaders since they can now more easily access the full information that was previously summarized in a terse alt text.


HaoZeke (commented):
Thanks for this @Bisaloo. Would it be possible to embed only a few lines? Currently this adds way too much to the length of the document which would likely be problematic from an accessibility standpoint.

HaoZeke (changes requested):
See previous comment :)

Bisaloo:
I'm still unsure of what would be the best way to do this. 

I went with saving only `head()` of the data. Alternatively, we could truncate the files before reading them with `echo = FALSE` to not distract the readers from the main point but I'm always wary of doing operations behind the scenes.

HaoZeke (approved):
LGTM now, thanks @Bisaloo.

PR #605, opened 2023-12-30
Make unzipping instructions platform-independent

### LINKED ISSUE #484: Fix platform-specific setup instructions
The [setup instructions](http://swcarpentry.github.io/r-novice-inflammation/setup.html) include the following line:
> If it’s not unzipped yet, double-click on it to unzip it. You should end up with a new folder called `data`.

Double-click to unzip is default behavior on Mac but not Windows ([official Windows directions](https://support.microsoft.com/en-ae/help/4028088/windows-zip-and-unzip-files)). I assume there's some variability on Linux. Directions should not be platform-specific for lessons designed for multiple platforms.

For consistency with the abbreviated nature of the remainder of the instructions, I suggest changing to:
>If it’s not unzipped yet, unzip it. There should now be a folder called `data` in the `r-novice-inflammation` folder.

If the remainder of the instructions were more detailed, it might make sense to link to platform-specific instructions for true beginners there (that may not know how to unzip) but in that case, all three platforms would need to be represented.

Bisaloo:
Fix #484

@ndporter, I went ahead and implement your suggestion from #484 (hence why you are marked as co-author of this commit). Please let me know if you'd do it any differently. Otherwise, this will be merged in the next couple of weeks.

### DIFF
--- a/learners/setup.md
+++ b/learners/setup.md
@@ -12,8 +12,7 @@ You also need to download some files to follow this lesson:
 1. Make a new folder in your Desktop called `r-novice-inflammation`.
 2. Download [r-novice-inflammation-data.zip](data/r-novice-inflammation-data.zip)
   and move the file to this folder.
-3. If it's not unzipped yet, double-click on it to unzip it. You should end up
-  with a new folder called `data`.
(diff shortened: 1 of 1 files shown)


isaac-jennings (approved):
Reviewed and concur with wording toward being concise, and system agnostic.

PR #609, opened 2024-01-03
consistent lowercase for filename

maneesha:
The second time the file `script.R` is referenced, it starts with a capital S.  For consistency as file names are case sensitive, this PR changes it to lowercase.

### DIFF
--- a/episodes/01-starting-with-data.Rmd
+++ b/episodes/01-starting-with-data.Rmd
@@ -180,7 +180,7 @@ Once a variable is created, we can use the variable name to refer to the value i
 
 <img src="fig/tag-variables.svg" alt="Variables as Tags" />
 
-To see the value of a variable, we can print it by typing the name of the variable and hitting <kbd>CTRL + Return</kbd> (or <kbd>CTRL + Enter</kbd>) while working in the `script.R` file in the editor which is recommended. If we are working in the console directly, we need to hit <kbd>Return</kbd> (or <kbd>Enter</kbd>).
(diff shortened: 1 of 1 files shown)


Bisaloo (commented):
Thanks! For future reference, the first occurrence is here:

https://github.com/swcarpentry/r-novice-inflammation/blob/5101d3f80007e40d92948ac6ccbbf959895c3f31/episodes/01-starting-with-data.Rmd#L73-L75

Should we also harmonize style (``"`script.R`"`` vs `` `script.R` ``) in this PR?

episodes/01-starting-with-data.Rmd
```diff
@@ -180,7 +180,7 @@ Once a variable is created, we can use the variable name to refer to the value i
 
 <img src="fig/tag-variables.svg" alt="Variables as Tags" />
 
-To see the value of a variable, we can print it by typing the name of the variable and hitting <kbd>CTRL + Return</kbd> (or <kbd>CTRL + Enter</kbd>) while working in the Script.R file in the editor which is recommended. If we are working in the console directly, we need to hit <kbd>Return</kbd> (or <kbd>Enter</kbd>).
+To see the value of a variable, we can print it by typing the name of the variable and hitting <kbd>CTRL + Return</kbd> (or <kbd>CTRL + Enter</kbd>) while working in the `script.R` file in the editor which is recommended. If we are working in the console directly, we need to hit <kbd>Return</kbd> (or <kbd>Enter</kbd>).
```
ErinBecker, line 183:
```suggestion
To see the value of a variable, we can print it by typing the name of the variable and hitting <kbd>CTRL + Return</kbd> (or <kbd>CTRL + Enter</kbd>) while working in the "`script.R`" file in the editor which is recommended. If we are working in the console directly, we need to hit <kbd>Return</kbd> (or <kbd>Enter</kbd>).
```

ErinBecker:
Thank you @maneesha for correcting this inconsistency. I've harmonized the quotation mark styling as recommended by @Bisaloo.

PR #611, opened 2024-01-03
don't call something easy

maneesha:
Changes language about writing a new similar function as easy.


Bisaloo (approved):

Bisaloo:
Very good point! Thanks!

Issue #650, opened 2025-06-02
Test new GitHub Action for containerised lesson builds?


tobyhodges:
### How could the content be improved?

Hi @swcarpentry/r-novice-inflammation-maintainers 👋

@froggleston has been developing a modified version of the GitHub Actions workflows we use to build and deploy lesson websites, which uses a container image to control the build environment. Based on his tests so far, it looks like this adjusted approach will make the builds more resistent to upstream changes, e.g. the semi-regular problems we encounter when a new version of R or a major dependency is released.

We are looking for an RMarkdown-based lesson where we could test the new workflows out. Would you be happy to give it a try here? Rob has [tested the updated version on his own fork](https://github.com/froggleston/r-novice-inflammation/actions/runs/15391479400) already and everything is looking good. 

We'd like to apply the changes here then monitor how things run for a few months, before (hopefully) rolling out the new workflows across all of The Carpentries' lesson repositories.

Bisaloo:
Sounds good to me! I'm excited to see this in action.

froggleston:
And I've [added pngquant](https://github.com/carpentries/workbench-docker/commit/dfd1820e8538d473da2c443c18d2a61749dc96b9) to the Dockerfile so will be in the next Workbench docker release.

froggleston:
So the scheduled build today failed, as I believe the cache expired just before the run was due to commence (caches are kept for 7 days). I've manually run Apply Package Cache to rebuild and cache again, and the build has succeeded.

froggleston:
Correct! We can monitor this next week.

froggleston:
The build seems to have succeeded fine this time round, due to when the caches were made.

Bisaloo:
I can also confirm it just worked great on a push.

froggleston:
I've opened #673 to improve some of the workflow behaviour and reduce complexity in the repo workflow files themselves.

Bisaloo:
I merged `main` into #676 and it's failing because it cannot access the `${{ secrets.AWS_GH_OIDC_REGION }}` secret here:

https://github.com/swcarpentry/r-novice-inflammation/blob/f593a57df293e68fefaf1554904b9d013b697deb/.github/workflows/docker_pr_receive.yaml#L195-L204

Log: https://github.com/swcarpentry/r-novice-inflammation/actions/runs/19857992368/job/56900706519?pr=676

froggleston:
[...]

PR #652, opened 2025-06-02
Manually update workflows to docker versions to test

froggleston:
This PR will replace the current Workbench workflows with dockerised versions to improve a range of maintenance factors, and deliver more reproducible builds.

Addresses https://github.com/swcarpentry/r-novice-inflammation/issues/650

This is still in testing phase, and can be safely reverted if issues arise.

### DIFF
--- a/.github/workflows/update-workflows.yaml
+++ b/.github/workflows/update-workflows.yaml
@@ -41,7 +41,7 @@ jobs:
 
       - name: Update Workflows
         id: update
-        uses: froggleston/actions/update-workflows@main
+        uses: carpentries/actions/update-workflows@main
         with:
           repo: ${{ github.event.inputs.tarball }} || "https://carpentries.r-universe.dev"
           clean: ${{ github.event.inputs.clean }} || ".yaml"


Bisaloo (commented):
Just a quick check to avoid additional commits, but you might have done this already:

I suspect the swcarpentry GH organization restricts [which actions can be used](https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#allowing-select-actions-and-reusable-workflows-to-run). 

Do you need to move `froggleston/actions` as a branch in `carpentries/actions` or enable actions from your account? (no extra commit required for this second option)

froggleston:
> Just a quick check to avoid additional commits, but you might have done this already:
> 
> I suspect the swcarpentry GH organization restricts [which actions can be used](https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#allowing-select-actions-and-reusable-workflows-to-run).
> 
> Do you need to move `froggleston/actions` as a branch in `carpentries/actions` or enable actions from your account? (no extra commit required for this second option)

I've merged in the changes to the main carpentries actions - they're non-breaking for existing workflows.

froggleston:
Thanks @Bisaloo ! Here we go .... 🤞

froggleston:
@Bisaloo Looks good! There were a couple of minor issues but none that stopped the build from working, just things I'll need to document and improve (the build is quite verbose as there are more checks now than before).

Bisaloo:
It's not really a problem in this specific case but might(?) be in other lessons: the png files created by rmarkdown are larger than what they use to be. I can't pinpoint the exact reason for this but my intuition would be:

- either pngquant used to be present and is no longer there. Knitr has a hook to automatically compress png as they are rendered if pngquant is installed and the hook is enabled.
- or the graphical backend is different. In particular, the `ragg` package probably produces png smaller than Cairo. I'm not sure which one was used before and which one is used now.

Again, this may not warrant a deep dive now but it's probably good to keep in the back of your mind.

froggleston:
That's great feedback - thank you! I can add pngquant to the docker image for the next release. I'm not doing anything with any internals, so haven't switched any backend rendering specifically.

PR #671, opened 2025-08-05
Try removing external function files

Bisaloo:


Bisaloo:
:thinking:, this doesn't seem to be the right fix since all these files are actually still in use.

It seems however some of them are executed while rendering the lesson (https://github.com/swcarpentry/r-novice-inflammation/commit/0a603c12c44f4573a4db34c6f0c6626cd09f61c7), which probably shouldn't happen.

I will keep thinking about this more on my end. Any suggestions on your end?

PR #673, opened 2025-09-30
Update docker workflows

froggleston:
This PR updates the sandpaper GHA workflows to new versions which:

- are smaller and cleaner, using new `carpentries/actions` instead
- have improved triggers and checks


froggleston:
@Bisaloo I've updated these workflows to fix a range of recent issues, but they currently require a test branch of the carpentries/actions. As a result if everything now works as it should, then we'll need to do another update to use the main actions once my changes have been tested and merged in.

PR #678, opened 2025-11-13
Clearer function name

bkmgit:
_If this pull request addresses an open issue on the repository, please add 'Closes #NN' below, where NN is the issue number._

N/A

_Please briefly summarise the changes made in the pull request, and the reason(s) for making these changes._

Use a more descriptive function name

_If any relevant discussions have taken place elsewhere, please provide links to these._

N/A

### DIFF
--- a/episodes/02-func-R.Rmd
+++ b/episodes/02-func-R.Rmd
@@ -411,14 +411,14 @@ rescale <- function(v) {
 
 ## Functions to Create Graphs
 
-Write a function called `analyze` that takes a filename as an argument and displays the three graphs produced in the [previous lesson][01] (average, min and max inflammation over time).
-`analyze("data/inflammation-01.csv")` should produce the graphs already shown, while `analyze("data/inflammation-02.csv")` should produce corresponding graphs for the second data set.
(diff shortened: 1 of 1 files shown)


HaoZeke (changes requested):
Thanks for the contribution.

I'm requesting changes rather than merging because the rename is incomplete against the rest of the lesson:

- `analyze` is used heavily after episode 02 (`03-loops-R`, `04-cond`, `15-supp-loops-in-depth`, `inflammation_fxns.R`, etc.). Renaming only the challenge/solution in `02-func-R.Rmd` to `plotme` would leave learners with a different name than every later episode.
- Even within `02-func-R.Rmd`, the hidden chunk around the advanced function still defines `analyze` (see the `challenge-more-advanced-function-analyze` block).

`plotme` is also a weaker pedagogical name than `analyze` for a function that loads data and plots summary statistics (mean/min/max over time).

If the goal is clearer naming, please either:

1. keep `analyze` here for consistency, or
2. rename **everywhere** the function appears across episodes (and update figure captions/call sites).

The "average" → "mean" wording in the challenge text is fine as a separate, smaller edit if you want to land that alone.

